PR #31, opened 2020-01-08
Allow `$pattern` to be set by `setPattern()` for FormMonthSelect

fabiang:
Hello,

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

somehow I don't see a reason why it shouldn't be possible to pass the pattern to the FormMonthSelect view helper.

When allowing this its possible to add separators between elements easily:

```
<div class="row">
    <div class="col-lg-7">
        <?=
            $this->formMonthSelect()
            ->setPattern('MMMM')
            ->render($element);
        ?>
    </div>
    <div class="col-lg-5">
        <?=
             $this->formMonthSelect()
            ->setPattern('y')
            ->render($element);
        ?>
    </div>
</div>
```

I don't like to add any test or documentation for this, as this is a very simple change.

### DIFF
--- a/docs/book/helper/form-month-select.md
+++ b/docs/book/helper/form-month-select.md
@@ -21,3 +21,30 @@ echo $this->formMonthSelect($monthYear);
 // <select name="monthyear[month]"> ... </select>
 // <select name="monthyear[year]"> ... </select>
 ```
+
+## Advanced usage
+
+### Render each element individually
+
+This can be archived by passing a custom pattern via `setPattern()`:
+
+```php
+<div class="row">
+    <div class="col-lg-7">
+        <!-- Render the select field for month only -->
+        <?=
+            $this->formMonthSelect()
+            ->setPattern('MMMM')
+            ->render($element);
+        ?>
+    </div>
+    <div class="col-lg-5">
+        <!-- Render the select field for day only -->
+        <?=
+             $this->formMonthSelect()
+            ->setPattern('y')
+            ->render($element);
+        ?>
+    </div>
+</div>
+```
--- a/src/View/Helper/FormMonthSelect.php
+++ b/src/View/Helper/FormMonthSelect.php
@@ -194,6 +194,16 @@ class FormMonthSelect extends AbstractHelper
     }
 
     /**
+     * @param string $pattern
+     * @return self
+     */
+    public function setPattern($pattern)
+    {
+        $this->pattern = $pattern;
+        return $this;
+    }
+
+    /**
      * Set date formatter
      *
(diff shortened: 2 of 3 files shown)


froschdesign:
@fabiang 
First: thank you for your time and this contribution!

> I don't like to add any test or documentation for this, as this is a very simple change.

Apart from the fact that a test is always required; without documentation nobody can know about the existence of this functionality. So it cannot be the goal that only the two of us have knowledge of it. 😉 
Therefore the documentation must also be extended.

fabiang:
@froschdesign Your wishes have become true :neckbeard:. PR updated.

docs/book/helper/form-month-select.md
```diff
@@ -21,3 +21,30 @@ echo $this->formMonthSelect($monthYear);
...
+            ->render($element);
+        ?>
+    </div>
+    <div class="col-lg-5">
+        <!-- Render the select field for day only -->
+        <?=
+             $this->formMonthSelect()
+            ->setPattern('y')
+            ->render($element);
+        ?>
```
froschdesign: I think, the real goal is a different one: you want to render the elements separately and not to set the pattern manually.

Therefore I suggest a solution which adds two new methods to allow the rendering of the month and year elements separately. This protects the internal processing of the pattern.

The underlying `FormSelect` helper allows already the rendering of parts:

https://github.com/laminas/laminas-form/blob/1f87e9f98874c7f82b8bf2f9b85c902c8d2dd04c/src/View/Helper/FormSelect.php#L177-L181
https://github.com/laminas/laminas-form/blob/1f87e9f98874c7f82b8bf2f9b85c902c8d2dd04c/src/View/Helper/FormSelect.php#L253-L255

The `FormCollection` helper has a method to render the template:

https://github.com/laminas/laminas-form/blob/1f87e9f98874c7f82b8bf2f9b85c902c8d2dd04c/src/View/Helper/FormCollection.php#L166-L172

fabiang: We could instead define a prefix, separator and suffix property (with getters and setters) which get added to the internal array which holds the elements which should be rendered.

froschdesign: Why so complicated? What's wrong with my proposal?

> I suggest a solution which adds two new methods to allow the rendering of the month and year elements separately.

weierophinney:
Closing due to inactivity. If you wish to re-open this pull request, please create a new pull request that addresses the feedback provided.

PR #55, opened 2020-01-16
Fix for PHP 7.4

remicollet:
```

1) LaminasTest\Form\FieldsetTest::testExtractOnAnEmptyTraversable
array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead

/work/GIT/laminas/laminas-form/src/Fieldset.php:394
/work/GIT/laminas/laminas-form/src/Form.php:952
/work/GIT/laminas/laminas-form/test/FieldsetTest.php:95

ERRORS!
Tests: 2084, Assertions: 3230, Errors: 1, Skipped: 41.

```

Also switch to phpunit v7 as v6 doesn't work properly with 7.4

### DIFF
--- a/src/Fieldset.php
+++ b/src/Fieldset.php
@@ -412,10 +412,30 @@ class Fieldset extends Element implements FieldsetInterface
                 }
             }
 
-            if ($valueExists) {
+            if ($valueExists
+                || (is_array($data) && array_key_exists($name, $data))
+                || ($data instanceof Traversable && $this->propertyExists($data, $name))
+            ) {
                 $elementOrFieldset->setValue($data[$name]);
             }
         }
+    }
+
+    /**
+     * Compatibility function to work with any Traversable.
+     *
+     * @see https://bugs.php.net/bug.php?id=79384
+     *
+     * @param string $name
+     * @return bool
+     */
+    private function propertyExists(Traversable $object, $name)
+    {
+        if (property_exists($object, $name)) {
+            return true;
+        }
+
+        return array_key_exists($name, iterator_to_array($object));
     }
 
     /**
--- a/src/Fieldset.php
+++ b/src/Fieldset.php
@@ -391,7 +391,7 @@ class Fieldset extends Element implements FieldsetInterface
         }
 
         foreach ($this->iterator as $name => $elementOrFieldset) {
-            $valueExists = array_key_exists($name, $data);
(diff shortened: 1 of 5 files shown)


src/Fieldset.php
```diff
@@ -391,7 +391,7 @@ public function populateValues($data)
         }
 
         foreach ($this->iterator as $name => $elementOrFieldset) {
-            $valueExists = array_key_exists($name, $data);
+            $valueExists = isset($data[$name]);
```
michalbundyra: `isset` is not the same as `array_key_exists` in case `$data[$name] === null`.

It is fine in line 397 because we expect there array or traversable
It is fine in line 403 because we skip null value there anyway

It can be an issue in line 415, when we accept null.

michalbundyra:
@remicollet I've updated your PR. Could you have a look, please?

michalbundyra:
@froschdesign I was fooled before by trying to access any Traversable as an array, but it is not true, as Traversable does not mean ArrayAccess. It is ok for ArrayObject but any Traversable cannot be used like that. I've changed the script again, updated the test and the easiest solution I can see is just to convert Traversable to an array (`iterator_to_array` function), and then use array as it was before.

src/Fieldset.php
```diff
@@ -382,11 +382,15 @@ public function prepareElement(FormInterface $form)
      */
     public function populateValues($data)
```
Xerkus: Does it really accept Traversable? Looks like docblock is wrong here

michalbundyra: As mentioned in my comments, it was not implemented correctly and tests were using just ArrayObject (so ArrayAccess), but for any custom Traversable it was not working. I've updated tests and align the code to work with Traversable.

We cannot remove Traversable  here as it would be BC break, because before we accept it with a bug, but accept it. Check other tests and condition of I've updated below.

src/Fieldset.php
```diff
@@ -382,11 +382,15 @@ public function prepareElement(FormInterface $form)
...
     {
-        if (! is_array($data) && ! $data instanceof Traversable) {
+        if ($data instanceof Traversable) {
+            $data = iterator_to_array($data);
```
michalbundyra: Just thinking if I should use here `ArrayUtils::iteratorToArray`...

michalbundyra:
Thanks, @remicollet!

PR #57, opened 2020-02-02
Update Element.php

WinterSilence:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

1. if we set `value` as attribute, then we can get it same, dont add `value` in other methods because it's not so usefull. it would be better to remove the `value` assignment here(in `setAttribute()`), but this would affect backward compatibility.
2. `return;` is void, not `null
3. fix `@return` and other omissions in comments

### DIFF
--- a/src/Element.php
+++ b/src/Element.php
@@ -104,7 +104,7 @@ class Element implements
     /**
      * Get value for name
      *
-     * @return string|int
+     * @return mixed
      */
     public function getName()
     {
@@ -195,28 +195,27 @@ class Element implements
      */
     public function setAttribute($key, $value)
     {
-        // Do not include the value in the list of attributes
-        if ($key === 'value') {
+        if ($key == 'value') {
+            // Do not include the value in the list of attributes
             $this->setValue($value);
-            return $this;
+        } else {
+            $this->attributes[$key] = $value;
         }
-        $this->attributes[$key] = $value;
         return $this;
     }
 
     /**
      * Retrieve a single element attribute
      *
-     * @param  $key
-     * @return mixed|null
+     * @param string $key
+     * @return mixed
      */
     public function getAttribute($key)
     {
-        if (! isset($this->attributes[$key])) {
-            return;
+        if ($key == 'value') {
+            return $this->value;
         }
-
-        return $this->attributes[$key];
+        return isset($this->attributes[$key]) ? $this->attributes[$key] : null;
     }
 
     /**
@@ -327,7 +326,7 @@ class Element implements
     /**
      * Set the label used for this element
      *
-     * @param $label
+     * @param  string $label
      * @return $this
      */
     public function setLabel($label)
@@ -447,16 +446,12 @@ class Element implements
     /**
      * Retrieve a single label option
      *
-     * @param  $key
-     * @return mixed|null
+     * @param  string $key
+     * @return mixed
      */
     public function getLabelOption($key)
     {
-        if (! isset($this->labelOptions[$key])) {
-            return null;
-        }
-
-        return $this->labelOptions[$key];
+        return isset($this->labelOptions[$key]) ? $this->labelOptions[$key] : null;
     }
 
     /**
--- a/src/Element.php
+++ b/src/Element.php
@@ -104,7 +104,7 @@ class Element implements
     /**
      * Get value for name
      *
-     * @return mixed
+     * @return string|int
      */
     public function getName()
     {
@@ -195,12 +195,12 @@ class Element implements
      */
     public function setAttribute($key, $value)
     {
-        if ($key == 'value') {
+        if ($key === 'value') {
             // Do not include the value in the list of attributes
             $this->setValue($value);
-        } else {
-            $this->attributes[$key] = $value;
+            return $this;
         }
+        $this->attributes[$key] = $value;
         return $this;
     }
 
@@ -212,7 +212,7 @@ class Element implements
      */
     public function getAttribute($key)
     {
-        if ($key == 'value') {
+        if ($key === 'value') {
             return $this->value;
         }
         return isset($this->attributes[$key]) ? $this->attributes[$key] : null;


WinterSilence:
strange fail in Travis CI. this PR cant generate this error

src/Element.php
```diff
@@ -85,7 +85,7 @@ public function init()
      * Set value for name
      *
      * @param  string $name
-     * @return Element|ElementInterface
+     * @return static
```
Xerkus: For the fluent interface there is a special return type:
```suggestion
     * @return $this
```

michalbundyra:
@WinterSilence 
Thanks for your contribution. I've rebased your branch and pushed changes.

But... I am honestly unsure about this "feature". I have found the source when this tricky condition has been added to setAttribute:
https://github.com/zendframework/zendframework/commit/437ff0b5ddb956a17072889e0a341733cd3b3187

I am leaving it open for now. Would be nice if you can provide more informations why you think it is needed. Also Unit tests will be required before merging.

WinterSilence:
@michalbundyra I can't adds anymore to PR description:  getter/setter- it's classic pair. I vote to  add value in attributes only for input tag, but I dont want break backward compatibility to ZF.

WinterSilence:
@michalbundyra add test for `setValue()` too? only for `ElementTest`? `zendframework/zendframework@437ff0b` - Why laminas deletes test for value setter/getter?

src/Element.php
```diff
@@ -195,28 +195,27 @@ public function setOption($key, $value)
...
-        if ($key === 'value') {
+        if ($key == 'value') {
+            // Do not include the value in the list of attributes
             $this->setValue($value);
-            return $this;
+        } else {
+            $this->attributes[$key] = $value;
         }
-        $this->attributes[$key] = $value;
         return $this;
```
weierophinney: Please revert all of the above changes.

- We prefer strict checks (`===`), as they ensure the type as well as the value.
- We prefer returning from within conditionals, as they allow us to remove `else` conditions (and thus improve readability by reducing indentation).

The changes above are unnecessary.

src/Element.php
```diff
@@ -104,7 +104,7 @@ public function setName($name)
     /**
      * Get value for name
      *
-     * @return string|int
+     * @return mixed
```
weierophinney: No. This should only ever be a string or an integer, as the value is used for array keys. Revert, please.

src/Element.php
```diff
@@ -195,28 +195,27 @@ public function setOption($key, $value)
...
         return $this;
     }
 
     /**
      * Retrieve a single element attribute
      *
-     * @param  $key
-     * @return mixed|null
+     * @param string $key
+     * @return mixed
```
weierophinney: With the changes you are proposing, this should definitely be `null|mixed` for the return value, as `null` is what is returned if the attribute cannot be found.

src/Element.php
```diff
@@ -195,28 +195,27 @@ public function setOption($key, $value)
...
      */
     public function getAttribute($key)
     {
-        if (! isset($this->attributes[$key])) {
-            return;
+        if ($key == 'value') {
```
weierophinney: Please us a strict check here (`===`).

WinterSilence: I can’t imagine a situation in which this would make sense, can you showing example?

src/Element.php
WinterSilence: @weierophinney i'm use `SplString`

weierophinney:
@michalbundyra and I have done some research to determine why `getAttribute()` has not allowed retrieving the value.

The code in `setAttribute()` was added here: https://github.com/zendframework/zendframework/commit/437ff0b5ddb956a17072889e0a341733cd3b3187

That commit was part of https://github.com/zendframework/zendframework/pull/1989

The point of allowing `setAttribute('value', $value)` was to act as a BC shim to the original API of the `Element` class, so that when you call `setAttributes()` with a `value` key included, it continues to work.

However, all current elements use the `setValue()`/ `getValue()` pairs, and, when defining an element via configuration, you can provide the value via the `value` key — and we recommend that instead of defining the value within the element attributes, as it ensures that everything works correctly for elements such as checkboxes, radios, CSRF, etc. where the value has to be treated differently (due to how the value is represented in HTML).

As such, adding value awareness to `getAttributes()` is not something we want to support at this time.

We **ARE** still interested in the changes you suggest for annotations. If you'd be willing to amend your patch to include only those (with changes per feedback we've already provided), we'll review again to merge.

WinterSilence:
@weierophinney ok

src/Element.php
michalbundyra: @WinterSilence Check this out:
https://stackoverflow.com/questions/2401478/why-is-faster-than-in-php
also http://www.phpbench.com/

WinterSilence: @michalbundyra 1. it's true only when `$key` not string. 2. you convert type in any case - here or before pass in method. L - ogic (:

weierophinney: Using anything other than a string for attribute names, either for storing the values or comparisons, is unlikely something we want to support. Usage of things like `SplString` or objects implementing `__toString()` for purposes such as this are pretty niche, particularly considering a standard use case is to use _configuration_ to seed your form elements.

So, while I understand your argument, in this context, it doesn't hold weight, and the decision to use strict comparisons is a reasonable one to make.

src/Element.php
weierophinney: You "resolved" the conversation without reverting to the original as requested.

WinterSilence:
@weierophinney you want return errors?

weierophinney:
> @weierophinney you want return errors?

Not sure what you're referencing; can  you elaborate, please?

froschdesign:
@WinterSilence 
The message still stands:

> We **ARE** still interested in the changes you suggest for annotations. If you'd be willing to amend your patch to include only those (with changes per feedback we've already provided), we'll review again to merge.

To obtain a smooth review for you pull request and to allow a later succession via commit history, please follow these suggestions:

* do not mix different topics in one pull request
* use a meaningful title for your pull request 
* write useful commit messages

Thanks in advance and we are looking forward to further contributions! 😃

PR #60, opened 2020-03-17
Fixes usage Traversable objects

michalbundyra:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

In many places we suppose to accept any Traversable object but unfortunately it was not working correctly. I've added test cases to cover changes and fixes all places I've found. These are mainly:
- several Elements and method `setOptions`
- Collection element - `setObject` and `populateValues`

In couple other places converted if statement to else-if - after we convert Traversable to an array there is no point to check again if the variable is an array.

### DIFF
--- a/src/Element/Checkbox.php
+++ b/src/Element/Checkbox.php
@@ -57,16 +57,16 @@ class Checkbox extends Element implements InputProviderInterface
     {
         parent::setOptions($options);
 
-        if (isset($options['use_hidden_element'])) {
-            $this->setUseHiddenElement($options['use_hidden_element']);
+        if (isset($this->options['use_hidden_element'])) {
+            $this->setUseHiddenElement($this->options['use_hidden_element']);
         }
 
-        if (isset($options['unchecked_value'])) {
-            $this->setUncheckedValue($options['unchecked_value']);
+        if (isset($this->options['unchecked_value'])) {
+            $this->setUncheckedValue($this->options['unchecked_value']);
         }
 
-        if (isset($options['checked_value'])) {
-            $this->setCheckedValue($options['checked_value']);
+        if (isset($this->options['checked_value'])) {
+            $this->setCheckedValue($this->options['checked_value']);
         }
 
         return $this;
--- a/src/Element/Collection.php
+++ b/src/Element/Collection.php
@@ -509,9 +509,7 @@ class Collection extends Fieldset
     {
         if ($this->object instanceof Traversable) {
             $this->object = ArrayUtils::iteratorToArray($this->object, false);
-        }
-
-        if (! is_array($this->object)) {
+        } elseif (! is_array($this->object)) {
             return [];
         }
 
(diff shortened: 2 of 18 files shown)


src/Element/Checkbox.php
```diff
@@ -57,16 +57,16 @@ public function setOptions($options)
     {
         parent::setOptions($options);
 
-        if (isset($options['use_hidden_element'])) {
-            $this->setUseHiddenElement($options['use_hidden_element']);
+        if (isset($this->options['use_hidden_element'])) {
```
Xerkus: This change is dangerous due to the conditional side effects:
If `$this->options['use_hidden_element']` is set outside of `setOptions()` then change would be unreliably applied depending on whether `setOptions()` was invoked or not.

src/Element/Collection.php
```diff
@@ -162,11 +161,13 @@ public function allowObjectBinding($object)
      */
     public function setObject($object)
     {
-        if (! is_array($object) && ! $object instanceof Traversable) {
+        if ($object instanceof Traversable) {
+            $object = iterator_to_array($object);
```
Xerkus:
```suggestion
            $object = ArrayUtils::iteratorToArray($object);
```

src/View/Helper/FormElementErrors.php
```diff
@@ -65,22 +65,23 @@ public function __invoke(ElementInterface $element = null, array $attributes = [
...
     {
         $messages = $element->getMessages();
-        if (empty($messages)) {
-            return '';
-        }
-
-        $messages = $messages instanceof Traversable ? iterator_to_array($messages) : $messages;
-        if (! is_array($messages)) {
+        if ($messages instanceof Traversable) {
+            $messages = iterator_to_array($messages);
```
Xerkus: Also use statement.
```suggestion
            $messages = ArrayUtils::iteratorToArray($messages);
```

froschdesign:
@michalbundyra 
Does the `Laminas\Form\LabelAwareTrait` need to be tested with `Traversable` for `setLabelOptions` method?

src/Element/Checkbox.php
Xerkus: Same goes for all similar changes

Xerkus: Disregard as `$this->options` is getting completely replaced by `parent::setOptions()` call rather than getting merged with previously provided options as I expected.

src/Element/Collection.php
michalbundyra: Nope, as `ArrayUtils::iteratorToArray` is recursive by default and at this point we don't want to replace it recursively - see tests.

Of course you could say to set 2nd param to false - but then it will just use `iterator_to_array` internally with additional redundant checks.

src/View/Helper/FormElementErrors.php
michalbundyra: Similar situation as above, there is no reason to change it, unless you can provide a valid test case. Please note that I kept the same code as before, just optimised the condition.

PR #64, opened 2020-04-23
Adds usage in a mvc application to documentation

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Enhancement        | yes

Some tasks for later:

* add links to installation, cookbook recipes like "Add custom form element" and "Add custom validator / filter"
* update description for laminas-cli

### DIFF
--- a/docs/book/application-integration/usage-in-a-laminas-mvc-application.md
+++ b/docs/book/application-integration/usage-in-a-laminas-mvc-application.md
@@ -46,19 +46,15 @@ class AlbumForm extends Form implements InputFilterProviderInterface
             [
                 'name'    => 'title',
                 'filters' => [
-                    [
-                        'name' => StripTags::class,
-                    ],
-                    [
-                        'name' => StringTrim::class,
-                    ],
+                    ['name' => StripTags::class],
+                    ['name' => StringTrim::class],
                 ],
                 'validators' => [
                     [
                         'name'    => StringLength::class,
                         'options' => [
                             'min' => 1,
-                            'max' => 120,
+                            'max' => 100,
                         ],
                     ],
                 ],
--- a/docs/book/application-integration/usage-in-a-laminas-mvc-application.md
+++ b/docs/book/application-integration/usage-in-a-laminas-mvc-application.md
@@ -1,0 +1,181 @@
+# Usage in a laminas-mvc Application
+
+The following example shows _one_ potential use case of laminas-form within
+a laminas-mvc based application. The example uses a module, a controller and the
+laminas-form element manager.
+
+The example is based on the [tutorial application](https://docs.laminas.dev/tutorials/getting-started/overview/)
+which builds an album inventory system.
+
+Before starting, make sure laminas-form is installed and configured.
+
+## Create Form
+
+Create a form as separate class, e.g. `module/Album/src/Form/AlbumForm.php`:
+
+```php
+namespace Album\Form;
+
+use Laminas\Form\Element\Text;
+use Laminas\Form\Form;
+
+class AlbumForm extends Form
+{
+    public function init()
+    {
+        // Title
+        $this->add(
+            [
+                'name'    => 'title',
+                'type'    => Text:class,
+                'options' => [
+                    'label' => 'Title',
+                ],
+            ]
+        );
+    
+        // …
+    }
+}
+```
+
+## Using Form
+
+### Create Controller
+
+Using the form in a controller, e.g.
+`module/Album/Controller/AlbumController.php`:
+
+```php
+namespace Album\Controller;
+
+use Album\Form\AlbumForm;
+use Laminas\Form\FormInterface;
+use Laminas\Mvc\Controller\AbstractActionController;
+
+class AlbumController extends AbstractActionController
+{
+    /** @var FormInterface */
+    private $form;
+    
+    public function __construct(FormInterface $form)
+    {
+        $this->form = $form;
+    }
+    
+    public function addAction()
+    {
+        // Set action attribute
+        $this->form->setAttribute(
+            'action',
+            $this->url()->fromRoute('album', ['action' => 'add'])
+        );
+
+        $variables = ['form' => $this->form];
+        
+        if (! $this->getRequest()->isPost()) {
+            return $variables;
+        }
+
+        // Validation
+        $this->form->setData($this->getRequest()->getPost());
+        if (! $this->form->isValid()) {
+            return $variables;
+        }
+    
+        // …
+
+        return $this->redirect()->toRoute('album', ['action' => 'add']);
+    }
+}
+```
+
+### Create Factory for Controller
+
+Fetch the `AlbumForm` from the form element manager in a factory,
+e.g. `src/Album/Controller/AlbumControllerFactory.php`:
+
+```php
+namespace Album\Controller;
+
+use Album\Form\AlbumForm;
+use Interop\Container\ContainerInterface;
+use Laminas\ServiceManager\Factory\FactoryInterface;
+use Laminas\ServiceManager\PluginManagerInterface;
+
+class AlbumControllerFactory implements FactoryInterface
+{
+    public function __invoke(
+        ContainerInterface $container,
+        $requestedName,
+        array $options = null
+    ) {
+        /** @var PluginManagerInterface $formElementManager */
+        $formElementManager = $container->get('FormElementManager');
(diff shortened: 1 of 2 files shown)


docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
+# Usage in a laminas-mvc Application
+
+The following example shows _one_ potential use case of laminas-form within
+a laminas-mvc based application. The example uses a module, a controller and the
+laminas-form element manager.
+
+The example is based on the [tutorial application](https://docs.laminas.dev/tutorials/getting-started/overview/)
+which builds an album inventory system.
+
+Before starting, make sure laminas-form is installed and configured.
```
michalbundyra, line 10: I'd like to see a link here - to "installation and configuration".

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+
+```php
+namespace Album\Form;
+
+use Laminas\Form\Element\Text;
+use Laminas\Form\Form;
+
+class AlbumForm extends Form
+{
+    public function init()
```
michalbundyra, line 24: minor: maybe we should consider adding `: void` RTH?

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+    {
+        // Title
+        $this->add(
+            [
```
michalbundyra, line 28: I would move it to the previous line, so we can reduce one indent:
```php
$this->add([
    'name'    => 'title',
    ...
]);
```

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+use Album\Form\AlbumForm;
+use Laminas\Form\FormInterface;
+use Laminas\Mvc\Controller\AbstractActionController;
+
+class AlbumController extends AbstractActionController
+{
+    /** @var FormInterface */
+    private $form;
+    
+    public function __construct(FormInterface $form)
```
michalbundyra, line 61: I'd prefer to have `AlbumForm` here, then we will be able to take advantages of zend-di and autowiring.

(yeah, with interface it will be still possible but additional configuration will be required)

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+
+        // Validation
+        $this->form->setData($this->getRequest()->getPost());
+        if (! $this->form->isValid()) {
+            return $variables;
+        }
+    
+        // …
+
+        return $this->redirect()->toRoute('album', ['action' => 'add']);
```
michalbundyra, line 88: I am a bit confused, it looks like a infinite loop - we redirect to the same method over and over again.
I might be missing something here.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+
+```php
+namespace Album\Controller;
+
+use Album\Form\AlbumForm;
+use Interop\Container\ContainerInterface;
+use Laminas\ServiceManager\Factory\FactoryInterface;
+use Laminas\ServiceManager\PluginManagerInterface;
+
+class AlbumControllerFactory implements FactoryInterface
```
michalbundyra, line 106: I am not sure if we have any convention, but I would go with a bit more generic case:
_invokable factory without the interface_. Then we can use PSR-11 container interface, and we do not need two other arguments in invoke (as these are unused anyway).

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+        $variables = ['form' => $this->form];
+        
+        if (! $this->getRequest()->isPost()) {
+            return $variables;
+        }
+
+        // Validation
+        $this->form->setData($this->getRequest()->getPost());
+        if (! $this->form->isValid()) {
+            return $variables;
```
michalbundyra, line 83: Should we return something more to the template here, so we know that the form has been submitted but it is not valid?

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: At the moment we have no page for this topic. If you remember, we wanted to use the theme for this: https://github.com/laminas/documentation-theme/blob/master/theme/pages/installation.html

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+
+Create a form as separate class, e.g. `module/Album/src/Form/AlbumForm.php`:
+
+```php
+namespace Album\Form;
+
+use Laminas\Form\Element\Text;
+use Laminas\Form\Form;
+
+class AlbumForm extends Form
```
samsonasik, line 22: how about implements `Zend\InputFilter\InputFilterProviderInterface`, since the form will be pulled from `FormElementManager` manager.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,181 @@
...
+    'controllers' => [
+        'factories' => [
+            // Add this line
+            Controller\AlbumController::class => Controller\AlbumControllerFactory::class,
+        ],
+    ],
+    // Add the following array
+    'form_elements' => [
+        'factories => [
+            Form\AlbumForm::class => InvokableFactory::class,
```
samsonasik, line 157: with pull from `FormElementManager`, no need to register it.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: Good point!

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: There is no other action in the controller. 😉

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: > I am not sure if we have any convention…

Nothing is written down, so no.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: We can use the flash messenger.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
samsonasik: how about use prg ?

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: The initial idea was to start with a very simple form but I added the input filter configuration for better illustration of the usage.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: This is a _basic_ example how laminas-form can be used in MVC application. For each complete example we have the tutorials.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -0,0 +1,198 @@
...
+> Custom form elements can also be used in this way.
+>
+> Additionally the form element manager calls the `init` method _after_
+> instantiating the form, ensuring all dependencies are fully injected
+> first.
+
+## Register Form and Controller
+
+If no separate factory is required for the form, then the form element manager
+will instantiating the form class. Otherwise the form must be registered.
```
froschdesign, line 160: > Otherwise the form must be registered.

A link must be added here when we have the description for the form element manager. (See #36)

michalbundyra:
Thanks, @froschdesign!

PR #66, opened 2020-05-07
Adds more links to description of MVC integration

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Enhancement        | yes

The descriptions for application integrations should only be bundles of individual descriptions, which should already exist in the components. Therefore links must be set for all steps.


michalbundyra:
Thanks, @froschdesign!

PR #68, opened 2020-06-18
Allow autocomplete attribute for hidden fields

MatthiasKuehneEllerhold:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

We have an application that uses JS-triggered page reloads to refresh data. We noticed that in Firefox 77.0.1 values in hidden fields were not updating. 
- Tests with Chrome 83 and FF 75 were successfull -> hidden field values are updating correctly. 
- Manually reloading the page with F5, Strg-R or Command-R shows the bad behaviour. 
- Using a page reload without cache (e. g. Strg-Shift-R) fixes the error.

In order to fix this we need to set "autocomplete=off" to all hidden fields in order to deny Firefox the autocompletion. 
This PR adds the ability to use the "autocomplete" attribute on a hidden field (validated by the FormHidden View Helper).

Bug-Report: https://bugzilla.mozilla.org/show_bug.cgi?id=520561

### DIFF
--- a/src/View/Helper/FormHidden.php
+++ b/src/View/Helper/FormHidden.php
@@ -23,6 +23,7 @@ class FormHidden extends FormInput
         'form'           => true,
         'type'           => true,
         'value'          => true,
+        'autocomplete'   => true,
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


samsonasik:
DCO is required, please check https://github.com/laminas/laminas-form/pull/68/checks?check_run_id=783306560

MatthiasKuehneEllerhold:
Updated the unit tests and added an (old?) mozilla bug-report.

MatthiasKuehneEllerhold:
Thanks!

PR #70, opened 2020-07-27
Added 2 tests for the DateSelect element around null value semantics

### LINKED ISSUE #54: Calling SetValue on Element/DateSelect with an empty string set's to todays date
If a DateSelect element has it's value set to an empty string, due to the behaviour of php's date time object; the element is set to today's date. See offending code:

https://github.com/zendframework/zend-form/blob/29a930f8a66cad97b727be8bad65b1f2e1d288f8/src/Element/DateSelect.php#L106

This may be considered desirable; however if the option create_empty_option is set to create a default empty option, this behaviour prevents the element from being set back to the empty option. 

The fix would be to test for an empty string in that conditional, but is a potential BC break

---

Originally posted by @carnage at https://github.com/zendframework/zend-form/issues/8

jroedel:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

<!--
Tell us about why this change is necessary:

- Are you providing a QA improvement (additional tests, CS fixes, etc.) that
  does not change behavior?

  Additional tests that don't change behavior. It wasn't clear what happens when a null is set to the DateSelect element. This creates tests that clarify this.

### DIFF
--- a/test/Element/DateSelectTest.php
+++ b/test/Element/DateSelectTest.php
@@ -115,4 +115,24 @@ class DateSelectTest extends TestCase
 
         $this->assertSame($element, $element->setValue('2014-01-01'));
     }
+    
+    public function testNullSetValueIsSemanticallyTodayWithoutEmptyOption()
+    {
+        $element  = new DateSelectElement('foo');
+        $element->setShouldCreateEmptyOption(false);
+        $now = new \DateTime();
+        $element->setValue(null);
+        $value = $element->getValue();
+        // the getValue() function returns the date in 'Y-m-d' format
+        $this->assertEquals($now->format('Y-m-d'), $value);
+    }
+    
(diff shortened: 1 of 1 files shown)


weierophinney:
You mention this is a BC break; can you elaborate, please?

jroedel:
I guess it's not technically a BC break since setting a null value was never well supported (3 warnings are currently being thrown), but it's conceivable that a small number of users are counting on today's date being set instead of a null value. 

Note that this "change" in functionality only would occur when the user sets the `create_empty_option`.

This PR only adds the tests, no functionality.

Slamdunk:
Manually merged in #118

PR #74, opened 2020-10-09
Fix documentation bug 

mushroot:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

- About the projects https://github.com/orgs/laminas/projects/8#card-36888172
this is my way to set value options from database for a select element

### DIFF
--- a/docs/book/application-integration/usage-in-a-laminas-mvc-application.md
+++ b/docs/book/application-integration/usage-in-a-laminas-mvc-application.md
@@ -38,15 +38,6 @@ class AlbumForm extends Form implements InputFilterProviderInterface
             ],
         ]);
 
-        // Category
-        $this->add([
-            'name'  => 'category',
-            'type'  => Select::class,
-            'options' => [
-                'label' => 'Category',
-            ],
-        ]);
-    
         // …
     }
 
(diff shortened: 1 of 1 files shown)


froschdesign:
@mushroot 
> this is my way to set value options from database for a select element

1. The description for application integration must contain only basic example. Everything else should be described separated. The plan is to create a "Cookbook" section like in [Mezzio](https://docs.mezzio.dev/mezzio/v3/cookbook/autowiring-routes-and-pipelines/).
2. In the controller is not the best idea because it means to duplicate code if the form is used multiple times. A [custom element](https://discourse.laminas.dev/t/handling-laminas-forms-with-dependencies-the-view-results/1343/12?u=froschdesign) can be used or the `Laminas\Db\Adapter\AdapterAwareTrait` and a delegator.

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -32,11 +32,20 @@ class AlbumForm extends Form implements InputFilterProviderInterface
         // Title
         $this->add([
             'name'    => 'title',
-            'type'    => Text:class,
+            'type'    => Text::class,
```
froschdesign, line 35: Good catch! 👍

froschdesign (changes requested):
The bugfix is good but everything else should be handled separately and therefore reverted. Thanks in advance!

docs/book/application-integration/usage-in-a-laminas-mvc-application.md
mushroot: ok, I reverted it.

froschdesign (approved):

Slamdunk:
Thank you

PR #75, opened 2020-10-21
Add PHP 8.0 support

### LINKED ISSUE #73: PHP 8.0 support
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary

To be prepared for the december release of PHP 8.0, this repository has some additional TODOs to be tested against the new major version.

In order to make this repository compatible, one has to follow these steps:

- [ ] Modify `composer.json` to provide support for PHP 8.0 by adding the constraint `~8.0.0`
- [ ] Modify `composer.json` to drop support for PHP less than 7.3
- [ ] Modify `composer.json` to implement phpunit 9.3 which supports PHP 7.3+
- [ ] Modify `.travis.yml` to ignore platform requirements when installing composer dependencies (simply add `--ignore-platform-reqs` to `COMPOSER_ARGS` env variable)
- [ ] Modify `.travis.yml` to add PHP 8.0 to the matrix (*NOTE*: Do not allow failures as PHP 8.0 has a feature freeze since 2020-08-04!)
- [ ] Modify source code in case there are incompatibilities with PHP 8.0

Slamdunk:
Close https://github.com/laminas/laminas-form/issues/73


Slamdunk:
The failing tests show an issue within `laminas-code`, I'll try to update that library first

Slamdunk:
Yup, `laminas-code` needs update for PHP 8.0: https://travis-ci.com/github/laminas/laminas-code/jobs/403158984

composer.json
```diff
@@ -25,38 +25,45 @@
...
-        "laminas/laminas-servicemanager": "^2.7.5 || ^3.0.3",
-        "laminas/laminas-session": "^2.8.1",
-        "laminas/laminas-text": "^2.6",
-        "laminas/laminas-validator": "^2.6",
-        "laminas/laminas-view": "^2.6.2",
-        "phpunit/phpunit": "^5.7.27 || ^6.5.14 || ^7.5.20"
+        "doctrine/annotations": "^1.10.4",
+        "laminas/laminas-cache": "^2.9.0",
+        "laminas/laminas-captcha": "^2.9.0",
+        "laminas/laminas-code": "dev-php_80_support",
```
Slamdunk: Waiting for https://github.com/laminas/laminas-code/pull/47 to get merged, converting this PR in draft to be sure gets not merged by error

test/Annotation/AnnotationBuilderTest.php
```diff
@@ -242,7 +241,7 @@ public function testAllowsComposingMultipleChildEntities()
      * @dataProvider provideOptionsAnnotationAndComposedObjectAnnotation
      * @param string $childName
      *
-     * @group 7108
+     * @group laminas7108
```
froschdesign: These changes are unnecessary because the (most) numbers are from the [old Zend Framework repository](https://github.com/zendframework/zendframework/pull/7108) and we have no naming scheme for groups.

Thanks in advance!

froschdesign (changes requested):

froschdesign:
@Slamdunk 
Great work! 👍

test/Annotation/AnnotationBuilderTest.php
Slamdunk: I made these changes because PHPUnit errored saying bare numbers are not accepted as group name.
But it seems I was wrong, since it's not erroring anymore now :|

Slamdunk: I was right: reverting all `@group laminas(\d+)` I get:

```
$ vendor/bin/phpunit 
PHP Fatal error:  Uncaught TypeError: strpos() expects parameter 1 to be string, int given in ./laminas-form/vendor/phpunit/phpunit/src/Framework/TestSuite.php:899
Stack trace:
#0 ./laminas-form/vendor/phpunit/phpunit/src/Framework/TestSuite.php(899): strpos()
#1 ./laminas-form/vendor/phpunit/phpunit/src/Framework/TestSuite.php(287): PHPUnit\Framework\TestSuite->containsOnlyVirtualGroups()
#2 ./laminas-form/vendor/phpunit/phpunit/src/Framework/TestSuite.php(365): PHPUnit\Framework\TestSuite->addTest()
#3 ./laminas-form/vendor/phpunit/phpunit/src/Framework/TestSuite.php(504): PHPUnit\Framework\TestSuite->addTestSuite()
#4 ./laminas-form/vendor/phpunit/phpunit/src/Framework/TestSuite.php(529): PHPUnit\Framework\TestSuite->addTestFile()
#5 ./laminas-form/vendor/phpunit/phpunit/src/TextUI/XmlConfiguration/TestSuite/TestSuiteMapper.php(50): PHPUnit\Framework\T in ./laminas-form/vendor/phpunit/phpunit/src/TextUI/Command.php on line 103
```

froschdesign: This does not change the fact that `laminas` is wrong for this groups. 😉 
I would use something general like `issue`.

Thanks for the recheck!

Slamdunk: Done

composer.json
weierophinney: @Slamdunk laminas-code 3.5.0 has been released, with laminas/laminas-code#47 in it; go ahead and update, and ping me when you want a review.

test/FormElementManagerFactoryTest.php
```diff
@@ -30,7 +33,7 @@ public function testFactoryReturnsPluginManager()
 
         if (method_exists($elements, 'configure')) {
             // laminas-servicemanager v3
-            $this->assertAttributeSame($container, 'creationContext', $elements);
+            // $this->assertAttributeSame($container, 'creationContext', $elements);
```
Slamdunk: @weierophinney should I wait for https://github.com/laminas/laminas-servicemanager/pull/57/files#diff-95e7bc706299f5873580bf786f6d5d08c3c8e92b2cb11e2df31f3fa6e3d9dacfR26-R29 to restore this assertion or can I go ahead without it?

Slamdunk:
In the latest commit https://github.com/laminas/laminas-form/pull/75/commits/7eb79bfcf33dd928ccb1f5b604bb91483648ecfc I've dropped support to `laminas-servicemanager < 3.4`, hope it's ok for you

froschdesign (approved):
LGTM 👍

boesing:
> In the latest commit [7eb79bf](https://github.com/laminas/laminas-form/commit/7eb79bfcf33dd928ccb1f5b604bb91483648ecfc) I've dropped support to `laminas-servicemanager < 3.4`, hope it's ok for you

You mean `laminas-code < 3.4`, don't you?

Slamdunk:
> You mean `laminas-code < 3.4`, don't you?

No, both, take a look at the split diff of `require-dev` section in `composer.json`:

image

src/FormElementManager.php
```diff
@@ -32,10 +33,8 @@
  *
  * Enforces that elements retrieved are instances of ElementInterface.
  */
-class FormElementManagerV3Polyfill extends AbstractPluginManager
```
rieschl: Isn't renaming a class a BC break?

src/FormElementManager/FormElementManagerTrait.php
```diff
@@ -1,121 +0,0 @@
...
-use function sprintf;
-
-/**
- * Trait providing common logic between FormElementManager implementations.
- *
- * Trait does not define properties, as the properties common between the
- * two versions are originally defined in their parent class, causing a
- * resolution conflict.
- */
-trait FormElementManagerTrait
```
rieschl, line 25: ... and a deletion is also a BC break?

src/FormElementManager.php
froschdesign: I do not see the name as a problem because an alias was created:

https://github.com/laminas/laminas-form/blob/a028e927ed1b07a98e43c22d5e7cac796f7a0e9d/autoload/formElementManagerPolyfill.php#L12-L18

But the polyfill classes and the related trait are not marked as internal and can be overwritten in userland code. For a smoothly transition we need a deprecation – in code and documentation.

@Slamdunk 

The following steps are needed:

* create a new class for `Laminas\Form\FormElementManager`
* add deprecation notices to the old trait und the old polyfill classes under the namespace `Laminas\Form\FormElementManager`
* add a migration guide to the documentation and add the deprecation infos (an example for deprecations can be found on [Mezzio](https://docs.mezzio.dev/mezzio/v2/reference/migration-to-v2-2/))

Slamdunk: I don't think we need a transition.

Yes: at raw level this class has never been marked as `@internal`, so it may appear as a BC Break.

But both `FormElementManagerV2Polyfill` and `FormElementManagerV3Polyfill` are never documented, type-hinted nor used as-is since they have always be used only once aliased:

https://github.com/laminas/laminas-form/blob/a028e927ed1b07a98e43c22d5e7cac796f7a0e9d/src/ConfigProvider.php#L42

https://github.com/laminas/laminas-form/blob/a028e927ed1b07a98e43c22d5e7cac796f7a0e9d/src/Factory.php#L92

To me it's just a bug related to the missing `@internal`, a best practice introduced recently; when those polyfills where created no one used that annotation.

froschdesign: > Yes: at raw level this class has never been marked as `@internal`, so it may appear as a BC Break.

Correct, the marking of the class is missing therefore the problem is in this component and cannot be at the user's expense.

> But both `FormElementManagerV2Polyfill` and `FormElementManagerV3Polyfill` are never documented…

Unfortunately we cannot rely on this fact because the entire documentation (code and docs) are not good for laminas-form.

> …type-hinted nor used as-is since they have always be used only once aliased…

Because of the class alias you will find the following use in applications:

```php
/** @var \Laminas\Form\FormElementManager|\Laminas\Form\FormElementManager\FormElementManagerV3Polyfill $formElementManager */
$formElementManager = $container->get('FormElementManager');
```

This will not break an application if the polyfill is removed but an IDE can highlight a deprecation notice.

> To me it's just a bug related to the missing `@internal`…

Yes it is a bug but not the fault of the user.

---

Please take the necessary steps to ensure that the rules of semantic versioning are followed. Thanks in advance!

rieschl: @Slamdunk generally I agree with you. it _should_ have been marked as `@internal`, but it wasn't. so you can assume, that some users use the class directly, e.g. myself 🙂
we extend this V3 polyfill with a dummy class to be able to type hint to a class and allowing auto-wiring.
For my projects it's not a problem, I can resolve such a case pretty quickly because I'm quite familiar with laminas code. but probably lots of consumers aren't and will be annoyed for having to fix a bc break in a minor version (even though if they extend such a weird polyfill class they have it coming to them 😋).
on the other side it doesn't hurt to keep the code and mark it as deprecated.
I was having the same problem in laminas-session, you can [look it up there](https://github.com/laminas/laminas-session/pull/35). TL;DR: I copied the polyfill code to the new class so that they aren't used anymore and marked them as deprecated.
The V2 polyfill was deleted because it can't be used anymore. it's not a bc break in that case.

froschdesign: > generally I agree with you.

I also agree on this. It would be great if the obsolete classes could simply be removed, but now the component must deal with the problem.

Slamdunk: >  I was having the same problem in laminas-session, you can [look it up there](https://github.com/laminas/laminas-session/pull/35). TL;DR: I copied the polyfill code to the new class so that they aren't used anymore and marked them as deprecated.
> The V2 polyfill was deleted because it can't be used anymore. it's not a bc break in that case.

Correct me if I'm wrong: by the same logic, stated that in this PR I've removed support for SMv2, I can rearrange the steps asked by @froschdesign in:

1. Keep `2.16` version of `FormElementManagerTrait` and mark it as `@deprected` and unmaintained
1. Delete `FormElementManagerV2Polyfill` as unusable anyway, like you did in laminas-session
1. Move `FormElementManagerV3Polyfill` to `FormElementManager` (already done)
1. Copy-paste `FormElementManagerTrait` content into `FormElementManager` with the purpose of simplify maintenance (already done)
1. Create a "new" `FormElementManagerV3Polyfill` as empty subclass of the new `FormElementManager`
1. Mark the "new" `FormElementManagerV3Polyfill` as `@deprecated`

This way we can:

1. Get rid of an old class
1. Get rid of `class_alias` burden
1. Refactor the code to be more easily maintainable
1. Keep BC

Right?

froschdesign: @Slamdunk 
Very good solution! 👍 
_(Much better than my short list.)_

Slamdunk: Done

.travis.yml
```diff
@@ -20,41 +20,28 @@ env:
...
-    - php: 7.3
+    - php: nightly
       env:
         - DEPS=lowest
-    - php: 7.3
-      env:
-        - DEPS=latest
-    - php: 7.4
+        - COMPOSER_ARGS="--no-interaction --ignore-platform-reqs"
+    - php: nightly
```
froschdesign: At the moment version `8.0.1-dev` is used on Travis therefore it should be replaced with `8.0`.

See: https://travis-ci.com/github/laminas/laminas-form/jobs/457287514#L211

Slamdunk: Done

froschdesign: @Slamdunk 
Please check Slack for new message. Thanks! 😃

src/FormElementManagerFactory.php
```diff
@@ -11,31 +11,22 @@
...
 use Laminas\ServiceManager\AbstractPluginManager;
 use Laminas\ServiceManager\Config;
-use Laminas\ServiceManager\FactoryInterface;
+use Laminas\ServiceManager\Factory\FactoryInterface;
 use Laminas\ServiceManager\ServiceLocatorInterface;
 
 use function is_array;
 use function method_exists;
 
 class FormElementManagerFactory implements FactoryInterface
```
boesing: This class is not final and thus extendable by endusers.

We should keep it as is even tho most of it is not used anymore. Marking methods as deprecated aswell as the `creationOptions` property would be the best way.

src/FormElementManagerFactory.php
```diff
@@ -11,31 +11,22 @@
 use Interop\Container\ContainerInterface;
 use Laminas\ServiceManager\AbstractPluginManager;
 use Laminas\ServiceManager\Config;
-use Laminas\ServiceManager\FactoryInterface;
+use Laminas\ServiceManager\Factory\FactoryInterface;
```
boesing: Changing the interface is a BC break aswell.
Unless we are bumping major version here, we should avoid such changes.

boesing (changes requested):
Overall, good addition here. Thanks so far!

Please keep the BC in mind and re-evaluate the changes in the `FormElementManagerFactory` to avoid bc breaks for end users.

src/FormElementManagerFactory.php
Slamdunk: Right: I've `@deprecated` attribute and method, and still removed what can be safely removed

src/FormElementManager.php
```diff
@@ -29,54 +29,94 @@
 use function sprintf;
 
 /**
- * laminas-servicemanager v2-compatible plugin manager implementation for form elements.
+ * laminas-servicemanager v3-compatible plugin manager implementation for form elements.
  *
  * Enforces that elements retrieved are instances of ElementInterface.
  */
-class FormElementManagerV2Polyfill extends AbstractPluginManager
+class FormElementManager extends AbstractPluginManager
```
boesing, line 36: Oh, didn't see that.
So you've renamed the v2 plugin manager instead of the v3 one.

Please diff with the v3 plugin manager from 2.15.0 to see whats missing in here.

boesing (changes requested):
Thanks for implementing BC!

Just one more thing to reconsider.

src/FormElementManager.php
Slamdunk: >  So you've renamed the v2 plugin manager instead of the v3 one.

No, I've renamed the v3 one, don't know why git tells otherwise. It already is BC, no API changes introduced except for `protected $shareByDefault = false;` which can't work anyway with v3:

```diff
$ git diff 2.16.x:src/FormElementManager/FormElementManagerV3Polyfill.php php_80_support:src/FormElementManager.php
diff --git a/src/FormElementManager/FormElementManagerV3Polyfill.php b/src/FormElementManager.php
index aaa8202d..e9af6a3a 100644
--- a/src/FormElementManager/FormElementManagerV3Polyfill.php
+++ b/src/FormElementManager.php
@@ -6,10 +6,11 @@
  * @license   https://github.com/laminas/laminas-form/blob/master/LICENSE.md New BSD License
  */

-namespace Laminas\Form\FormElementManager;
+namespace Laminas\Form;

 use Interop\Container\ContainerInterface;
 use Laminas\Form\Element;
+use Laminas\Form\Exception;
 use Laminas\Form\ElementFactory;
 use Laminas\Form\ElementInterface;
 use Laminas\Form\Fieldset;
@@ -32,10 +33,8 @@ use function sprintf;
  *
  * Enforces that elements retrieved are instances of ElementInterface.
  */
-class FormElementManagerV3Polyfill extends AbstractPluginManager
+class FormElementManager extends AbstractPluginManager
 {
-    use FormElementManagerTrait;
-
     /**
      * Aliases for default set of helpers
      *
@@ -277,13 +276,6 @@ class FormElementManagerV3Polyfill extends AbstractPluginManager
      */
     protected $sharedByDefault = false;

-    /**
-     * Don't share form elements by default (v2)
-     *
-     * @var bool
-     */
-    protected $shareByDefault = false;
-
     /**
      * Interface all plugins managed by this class must implement.
      * @var string
@@ -399,4 +391,62 @@ class FormElementManagerV3Polyfill extends AbstractPluginManager

         return $this;
     }
+    /**
+     * Try to pull hydrator from the creation context, or instantiates it from its name
+     *
+     * @param  string $hydratorName
+     * @return mixed
+     * @throws Exception\DomainException
+     */
+    public function getHydratorFromName($hydratorName)
+    {
+        $services = $this->creationContext;
+
+        if ($services && $services->has('HydratorManager')) {
+            $hydrators = $services->get('HydratorManager');
+            if ($hydrators->has($hydratorName)) {
+                return $hydrators->get($hydratorName);
+            }
+        }
+
+        if ($services && $services->has($hydratorName)) {
+            return $services->get($hydratorName);
+        }
+
+        if (! class_exists($hydratorName)) {
+            throw new Exception\DomainException(sprintf(
+                'Expects string hydrator name to be a valid class name; received "%s"',
+                $hydratorName
+            ));
+        }
+
+        $hydrator = new $hydratorName;
+        return $hydrator;
+    }
+
+    /**
+     * Try to pull factory from the creation context, or instantiates it from its name
+     *
+     * @param  string $factoryName
+     * @return mixed
+     * @throws Exception\DomainException
+     */
+    public function getFactoryFromName($factoryName)
+    {
+        $services = $this->creationContext;
+
+        if ($services && $services->has($factoryName)) {
+            return $services->get($factoryName);
+        }
+
+        if (! class_exists($factoryName)) {
+            throw new Exception\DomainException(sprintf(
+                'Expects string factory name to be a valid class name; received "%s"',
+                $factoryName
+            ));
+        }
+
+        $factory = new $factoryName;
+        return $factory;
+    }
 }
```

boesing: hm, as I've looked into it yesterday, it looked like there were missing some aliases 🤔
I'll check that again when I find spare time. 
Thanks for re-checking.

weierophinney (approved):

src/FormElementManager.php
Slamdunk: @boesing may I help further on this?

Slamdunk:
So, `laminas-code:v4` has been released dropping support for Annostation scanning.

Libraries that rely on `laminas-form` that want to support PHP 8 are basically forced to upgrade to `laminas-code:v4`, since real test cases simply don't work on PHP 8, see for example https://github.com/doctrine/DoctrineORMModule/pull/650/checks?check_run_id=1692052713

The only solution I see to this is to upgrade `laminas-form` to `laminas-code:v4` too, which means a BC break, hence a new MAJOR, because API like https://github.com/laminas/laminas-form/blob/2.16.x/src/Annotation/AnnotationBuilder.php#L118 are going to break.

I'm going to work on this, but I'm not sure I can replace Annotation parsing.

Slamdunk:
In order to not reinvent the wheel, I'll try to use https://github.com/doctrine/annotations

Slamdunk:
After a day thinking about it, the issues affecting PHP 8 and `laminas-form` are only related to annotations, which is not a mandatory dependency.

So we can ship this as-is, declaring partial support for annotations on PHP 8. and then we can focus on the next MAJOR with a replacement for `laminas-code`'s annotations

charma:
> In order to not reinvent the wheel, I'll try to use https://github.com/doctrine/annotations

How about using php Attributes directly instead? Maybe this is an sufficiend alternative without the need of another dependency?
https://www.php.net/manual/de/language.attributes.overview.php

froschdesign:
@charma 
> How about using php Attributes directly instead? Maybe this is an sufficiend alternative without the need of another dependency?

Please open a new issue report for that topic. Thanks in advance! 👍

driehle:
I had a look into `doctrine/annotations`. I am quite confident that switching to that library will work. The structure of the annotations can stay the same, which I'd say is the most important thing. However, the signature of the annotations classes has to change slightly, for instance:

An annotation `@Required(true)` becomes:
- with `laminas/laminas-code`: `new Required(['value' => true])`
- with `doctrine/annotations`: `new Required(true)`

Hence, the constructors of all annotations classes have to be changed which might be considered a BC break, since those classes have not been marked as `@internal`. Maybe we could change the constructors from

```php
    public function __construct(array $data)
    {
        [...]
    }
```

to the following:

```php
    public function __construct($data)
    {
        if (!is_array($data) && !isset($data['value'])) {
            $data = ['value' => $data];
        }

        [...]
    }
```

Still, in the unlikely event that someone was extended the annotation classes and overwritten the constructor, things will likely break, as with `doctrine/annotations` the constructors of the annotation classes are simply invoked differently.

In consequence, it might be easier to release a new major (i.e. 3.0) with annotations support through `doctrine/annotations`. Additionally, support for PHP 8 attributes can be added later with a new minor (e.g. 3.1). What do you think?

boesing:
I wonder if its just possible to use `doctrine/annotations` `DocParser` to Interpret the annotations but instantiating with the old "API" somehow?
This way, we could avoid BC breaks.

Slamdunk:
Ok, I've verified again this PR against https://github.com/doctrine/DoctrineORMModule/pull/661 and all tests pass.

I consider this ready-to-merge as users can run this on PHP 8.0 without BC Breaks, but of course we need to deprecate the current Annotation stuff. We can postpone that until we find a proper alternative.

PR #79, opened 2021-02-02
fix Form::getInputFilter() phpdoc return type to match FormInterface

pine3ree:
https://github.com/laminas/laminas-form/blob/a028e927ed1b07a98e43c22d5e7cac796f7a0e9d/src/Form.php#L708 makes sure it is created.


Slamdunk:
Merged in https://github.com/laminas/laminas-form/pull/89 due to wrong base branch, thank you @pine3ree

PR #93, opened 2021-03-21
fix: revert bc-break introduced in version 2.16.0

### LINKED ISSUE #92: 2.16.0 removes FormElementManager::get()
### BC Break Report

|    Q        |   A
|------------ | ------
| Version     | 2.16.0

#### Summary

When refactoring the `FormElementManager` class in #75 and removing support for service-manager v2 the V3Polyfill and `FormElementManagerTrait` were copied together to a "real" FormElementManager class.
But it seems that the previous method `\Laminas\Form\FormElementManager\FormElementManagerTrait::get` wasn't copied.
The main purpose of that method was calling `\Laminas\Form\FormElementManager::setInvokableClass()` for invokable Forms to override the default InovkableFactory with `\Laminas\Form\ElementFactory`.

#### Previous behavior

Invokable Forms (and probably all other Elements and Fieldsets) were created with the `\Laminas\Form\ElementFactory`.

#### Current behavior

Invokable Forms with options are instantiated incorrectly, because `\Laminas\ServiceManager\Factory\InvokableFactory` instantiates the object differently.

#### How to reproduce

Fetch a Form with options from the FormElementManager
```
$this->formElementManager->get(Form::class, ['foo' => 'bar']);
```
Previously, when no custom factory was registered, `FormElementManagerTrait::get` called `FormElementManager::setInvokableClass` to use `ElementFactory`, which creates the object with `return new $requestedName($name, $options);`. Now, this is missing, causing `InvokableFactory` to instantiate the class with `new $requestedName($options)`, using the options array as form name.
The default constructor for `Fieldset` which `Form` extends from, is `public function __construct($name = null, $options = [])`.
The options array is used as form name and later on, when rendering the form, the [escaper view helper](https://github.com/laminas/laminas-view/blob/2.12.x/src/Helper/Escaper/AbstractHelper.php#L55) throws up when trying to escape the `name` atrribute of the form opening tag because it's an array and not a string.

rieschl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Re-adds `\Laminas\Form\FormElementManager\FormElementManagerTrait::get` method to `\Laminas\Form\FormElementManager` to restore previous behavior.

fixes #92

### DIFF
--- a/src/FormElementManager.php
+++ b/src/FormElementManager.php
@@ -9,13 +9,7 @@
 namespace Laminas\Form;
 
 use Interop\Container\ContainerInterface;
-use Laminas\Form\Element;
 use Laminas\Form\Exception;
-use Laminas\Form\ElementFactory;
-use Laminas\Form\ElementInterface;
-use Laminas\Form\Fieldset;
-use Laminas\Form\Form;
-use Laminas\Form\FormFactoryAwareInterface;
 use Laminas\ServiceManager\AbstractPluginManager;
 use Laminas\ServiceManager\Exception\InvalidServiceException;
 use Laminas\Stdlib\InitializableInterface;
@@ -23,9 +17,11 @@ use Laminas\Stdlib\InitializableInterface;
 use function array_push;
 use function array_search;
 use function array_unshift;
+use function class_exists;
 use function get_class;
 use function gettype;
 use function is_object;
+use function is_string;
 use function sprintf;
 
 /**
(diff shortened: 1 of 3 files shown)


Slamdunk:
Thank you very much @rieschl

PR #98, opened 2021-03-23
Adds new method for fieldset class to set hydrator by name

froschdesign:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

# Description

## Current Usage

In a form or fieldset, a custom hydrator can be set via the form factory. The form factory uses the form element manager to fetch the hydrator from `HydratorManager` or the service container.
A separate factory for the form itself is then not required.

```php
class ExampleForm extends Laminas\Form\Form
{
    public function init(): void
    {
        $formElementManager = $this->getFormFactory()->getFormElementManager();
        $hydrator           = $formElementManager->getHydratorFromName(
            ExampleHydrator::class
        );
        $this->setHydrator($hydrator);
    }
}
```

https://github.com/laminas/laminas-form/blob/97a7da80bc0ab3287884235d89f4460fe2a170a5/src/FormElementManager.php#L424-L455

The same way via form factory and form element manager is already used for elements:

```php
class ExampleForm extends Laminas\Form\Form
{
    public function init(): void
    {
        $this->add(['type' => ExampleElement::class]);
    }
}
```

https://github.com/laminas/laminas-form/blob/97a7da80bc0ab3287884235d89f4460fe2a170a5/src/Fieldset.php#L161-L166
https://github.com/laminas/laminas-form/blob/97a7da80bc0ab3287884235d89f4460fe2a170a5/src/Factory.php#L123-L128

See also the [creation via factory](https://docs.laminas.dev/laminas-form/quick-start/#creation-via-factory).

## New Proposed Method

To simplify the handling for the end user, I propose a new method that shortens the setting of the hydrator. This also simplifies documentation and the associated explanation.

```php
class ExampleForm extends Laminas\Form\Form
{
    public function init(): void
    {
        $this->setHydratorByName(ExampleHydrator::class);
    }
}
```
A new method is used here because an interface cannot simply be changed.

https://github.com/laminas/laminas-form/blob/97a7da80bc0ab3287884235d89f4460fe2a170a5/src/FieldsetInterface.php#L120-L126

### DIFF
--- a/src/Fieldset.php
+++ b/src/Fieldset.php
@@ -476,6 +476,24 @@ class Fieldset extends Element implements FieldsetInterface
     }
 
     /**
+     * Set the hydrator by name to use when binding an object to the element.
+     *
+     * The form element manager {@see FormElementManager} is used via the
+     * form factory {@see Factory} to fetch the hydrator.
+     *
+     * @throws Exception\DomainException If hydrator is not found in hydrator
+     *                                   manager or service manager.
+     */
+    public function setHydratorByName(string $hydratorName): void
+    {
+        $this->setHydrator(
+            $this->getFormFactory()
+                ->getFormElementManager()
+                ->getHydratorFromName($hydratorName)
+        );
+    }
+
+    /**
      * Get the hydrator used when binding an object to the fieldset
      *
      * If no hydrator is present and object implements HydratorAwareInterface,
--- a/src/Form.php
+++ b/src/Form.php
@@ -925,4 +925,13 @@ class Form extends Fieldset implements FormInterface
 
         return $values;
     }
+
+    public function setInputFilterByName(string $inputFilterName): void
+    {
+        $inputFilter = $this->getFormFactory()
+            ->getInputFilterFactory()
+            ->getInputFilterManager()
+            ->get($inputFilterName);
+        $this->setInputFilter($inputFilter);
+    }
 }
--- a/src/Form.php
+++ b/src/Form.php
@@ -926,6 +926,13 @@ class Form extends Fieldset implements FormInterface
         return $values;
     }
 
+    /**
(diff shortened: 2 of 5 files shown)


Slamdunk:
Hi, this seems good to me, why it's still in draft?

froschdesign:
@Slamdunk 
> …why it's still in draft?

I wanted to get some feedback first, because the next step would be to add the same function for input filter:

```php
$inputFilter = $this->getFormFactory()
    ->getInputFilterFactory()
    ->getInputFilterManager()
    ->get(ExampleInputFilter::class);
$this->setInputFilter($inputFilter);
```

Slamdunk:
I see you didn't edit the interfaces, and that's good to me since this is just a shortcut and not needed as API.

So to me it's a good to go :+1:

froschdesign:
> …and that's good to me since this is just a shortcut and not needed as API.

Thank you, that's exactly what I wanted to hear / read. 😃

Slamdunk:
@froschdesign do you want me to finish this PR?

froschdesign:
@Slamdunk 
That would be great. Thanks in advance! ❤️

Slamdunk:
@froschdesign I have no idea where to put the doc for this:

* `creation-via-factory.md` and `using-annotations.md` have no interest for this feature
* `factory-backed-form-extension.md` should be the right place, but it only make sense in the circumstance of having a configuration that do something to the `Hydrator`/`InputFilter` that you can't do by just instantiating them inline, and so a more comprehensive example with configuration would be needed; a lot of work :\
* `programmatic-form-creation.md` may be interested, may be not

froschdesign:
> …and so a more comprehensive example with configuration would be needed; a lot of work

I will take care of that. I will create a separate pull request, so feel free to merge this one.

Issue #100, opened 2021-04-07
Class 'Laminas\Code\Annotation\AnnotationManager' not found


dbierer:
### Bug Report

|    Q        |   A
|------------ | ------
| laminas/laminas-form | 2.16.2

#### Summary

Attempting to build a form using Laminas\Form\Annotation\Annotation\Builder

#### Current behavior

```
root@laminas [ /home/onlinemarket.complete ]# php annotation_test.php 
PHP Fatal error:  Uncaught Error: Class 'Laminas\Code\Annotation\AnnotationManager' not found in /home/onlinemarket.complete/vendor/laminas/laminas-form/src/Annotation/AnnotationBuilder.php:178
Stack trace:
#0 /home/onlinemarket.complete/vendor/laminas/laminas-form/src/Annotation/AnnotationBuilder.php(221): Laminas\Form\Annotation\AnnotationBuilder->getAnnotationManager()
#1 /home/onlinemarket.complete/vendor/laminas/laminas-form/src/Annotation/AnnotationBuilder.php(257): Laminas\Form\Annotation\AnnotationBuilder->getFormSpecification()
#2 /home/onlinemarket.complete/annotation_test.php(19): Laminas\Form\Annotation\AnnotationBuilder->createForm()
#3 {main}
  thrown in /home/onlinemarket.complete/vendor/laminas/laminas-form/src/Annotation/AnnotationBuilder.php on line 178
```

#### How to reproduce

1. `composer create-project -sdev laminas/laminas-mvc-skeleton path/to/install`
2. `cd path/to/install`
3. `composer require laminas/laminas-form`
4. `composer require laminas/laminas-code`
5. `php annotation_test.php`
```
<?php
// annotation_test.php
require __DIR__ . '/vendor/autoload.php';
use Laminas\Hydrator\ClassMethodsHydrator;
use Laminas\Form\Annotation\AnnotationBuilder;
/**
 * @ABC\Name("attendee")
 * @ABC\Hydrator("Laminas\Hydrator\ClassMethodsHydrator")
 */
class AttendeeEntity {
    /**
    * @ABC\Name("name")
    * @ABC\Attributes({"type":"text","placeholder":"Name on Ticket"})
    * @ABC\Options({"label":"Name:"})
    * @ABC\Filter({"name":"StringTrim"})
    * @ABC\Filter({"name":"StripTags"})
    */
    protected $name;
}
$form = (new AnnotationBuilder())->createForm('AttendeeEntity');
[...]

PR #101, opened 2021-04-07
Conflict with laminas-code >= 4 to ensure BC

Slamdunk:
Fixes https://github.com/laminas/laminas-form/issues/100

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "laminas/laminas-zendframework-bridge": "^1.1"
     },
     "conflict": {
-        "laminas/laminas-code": "< 3.5.0"
+        "laminas/laminas-code": "<3.5.0 || >=4.0.0"
     },
     "require-dev": {
         "doctrine/annotations": "^1.10.4",


Ocramius:
Note that this will only prevent `2.16.3` to be installed, while `2.16.2` is still installable

Slamdunk:
Sure, but at least I can say `2.16.3` "fixes" the issue :|
The same goes for https://github.com/laminas/laminas-form/pull/96
I had no other arrows on my quiver

Ocramius:
@Slamdunk you did all that is possible, no worries: I was mostly leaving a bradcrumb for people encountering this :-)

driehle:
@Slamdunk @Ocramius I disagree that this solves #100. For now, if your application requires laminas-code ^4.0 for some other reasons, compose will simply downgrade laminas-form to 2.16.2. That cannot be a solution, can it? I'd agree that this would be a solution if you'd remove 2.16.1 and 2.16.2 from packagist.

Moreover, I am am concerned if adding a conflict to laminas-code >= 4.0 is desirable at all. I'm having the issue that doctrine/migrations has a dependency (via friendsofphp/proxy-manager-lts) to laminas-code. So for all people that do not need laminas-code with laminas-form (because they don't use annotations), but who require laminas-code for other reasons, you are blocking the upgrade path.

Ocramius:
@driehle it does what's possible: that's sadly how i works with too permissive dependency ranges.

That's what https://github.com/laminas/laminas-form/pull/101#issuecomment-814743995 was about.

The issue is indeed solved to the best extent we could handle it.

Alternatively, a better fix is to release `2.17.0` with support for `^4` too, and a wider dependency range.

As for `friendsofphp/proxy-manager-lts`, I cannot speak for that: their maintenance policy is questionable from my PoV.

driehle:
@Ocramius see my proposal in #104 for supporting `laminas-code:^4.0` :-)

PR #104, opened 2021-04-23
Added support for annotations using doctrine/annotations

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no (unless some specific edge cases, see below)
| New Feature   | yes (updates laminas-code to 4.0)
| RFC           | yes
| QA            | no

### Description

As discussed in #101, the current conflicts constraint regarding `laminas-code` brings some troubles, especially because earlier `2.16.x` versions were released without that constraint (though not different in functionality). This leads to composer downgrading from `2.16.3` to earlier versions, which then will still trigger the errors mentioned in #100, as `2.16.0`, `2.16.1` and `2.16.2` can be installed together with `laminas-code:4.0`, which does not have support for annotations anymore.

Therefore, to keep support for annotations, a new solution has to be found. Since `laminas-code:3.x` was relying on `doctrine/annotations`, for me it makes sense to directly implement that functionality in `laminas-form`. This is what this PR provides.

However, there are a few caveats:
 - Since the classes `AnnotationManager` and `DoctrineAnnotationParser` do not exist in `laminas-code:4.0` anymore, the getters and setters in `AnnotationBuilder` cannot work anymore either. The signature of this functions has changed slightly, which technically is a BC break. However, that will only affect people who have extended `AnnotationBuilder` and overridden these functions.
 - Similarily, the class `AnnotationCollection` from `laminas-code` doesn't exist anymore. This class has been added to `laminas-form` now, in order to keep the same interfaces for working with collections of assertions. People who type-checked this collection will need to update their code.
 - In contrast to `laminas-code:3.x`, `doctrine/annotations` parses all annotations and does not limit itself to a white list of annotation classes. Consequently, an injection of custom annotations is not needed anymore and the related test cases has been removed. This is not likely to affect any people, as the configuration keys for injecting such custom annotations is simply ignored and custom annotations will still work via autoloading.

Let me know what you thing. :-)

### DIFF
--- a/src/Annotation/AnnotationBuilder.php
+++ b/src/Annotation/AnnotationBuilder.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-form for the canonical source repository
- * @copyright https://github.com/laminas/laminas-form/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-form/blob/master/LICENSE.md New BSD License
- */
-
 namespace Laminas\Form\Annotation;
 
 use ArrayObject;
--- a/src/Annotation/AbstractAnnotationsListener.php
+++ b/src/Annotation/AbstractAnnotationsListener.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-form for the canonical source repository
- * @copyright https://github.com/laminas/laminas-form/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-form/blob/master/LICENSE.md New BSD License
- */
-
 namespace Laminas\Form\Annotation;
 
 use Laminas\EventManager\AbstractListenerAggregate;
--- a/src/Annotation/AbstractArrayAnnotation.php
+++ b/src/Annotation/AbstractArrayAnnotation.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-form for the canonical source repository
- * @copyright https://github.com/laminas/laminas-form/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-form/blob/master/LICENSE.md New BSD License
- */
-
 namespace Laminas\Form\Annotation;
 
 use Laminas\Form\Exception;
--- a/src/Annotation/AbstractArrayOrStringAnnotation.php
+++ b/src/Annotation/AbstractArrayOrStringAnnotation.php
@@ -1,11 +1,5 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-form for the canonical source repository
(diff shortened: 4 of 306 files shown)


Slamdunk:
Hi, thank you very much for this PR.
Using doctrine annotations was my intention too, but as a BC break we need to release it as a new major, and that's ok for me.
I hope to review this soon

driehle:
Thanks, yeah, strictly speaking a BC break cannot be avoided because `laminas-code` made a BC break in `4.0` and `laminas-form` exposes interfaces from `laminas-code` that were affected (i.e. got removed) in the `laminas-code` major release.

However, if we go for a new major in `laminas-form`, shouldn't we directly change all interfaces to be typed?

src/Annotation/AnnotationBuilder.php
```diff
@@ -112,18 +114,16 @@ public function setFormFactory(Factory $formFactory)
...
     {
-        $parser = $this->getAnnotationParser();
-        foreach ($this->defaultAnnotations as $annotationName) {
-            $class = __NAMESPACE__ . '\\' . $annotationName;
-            $parser->registerAnnotation($class);
-        }
-        $annotationManager->attach($parser);
-        $this->annotationManager = $annotationManager;
+        trigger_error(sprintf(
+            'Usage of %s is deprecated since v2.17.0, as laminas-code 4.0 does not have an annotation manager anymore.',
```
Slamdunk: @Ocramius It seems to me that having a `v2.17` release with this change is both a BC-Break and a pita, so I'd like to go all-in and jump directly into a `v3` release with all this removed in favor of `doctrine/annotations`: wdyt?
/cc @froschdesign

Ocramius: To me, yes, adding a warning/deprecation is problematic.

Not my call though, as I'm not involved enough in `laminas-form` to understand the next milestones.

froschdesign: The warning / deprecation notice would be okay if the feature is still there, then the user can suppress the notice and everything works like before.
But at this point the notice comes too late therefore we must go straight to the next step.

driehle: It might still be reasonable to release a 2.17.0 (or 2.16.4) with the deprecation notice (and the old behaviour, i.e. laminas-code 3.5.0) and then in parallel make a 3.0 release with all the new stuff. I can provide two PRs if you like. Please let me know if the deprecation notice should be based on 2.16.x or 2.17.x branch.

Slamdunk: We can't because the class isn't `final` and so we cannot avoid the BC-Break.

Btw all the blame should go to @Ocramius who released a `laminas-code` new major with a tenth of the functionalities of the previous one, and so I'm therefore allowed to do big jumps here too :tongue:

Ocramius: "Soft" dependencies, y'all - they come from an age where we were very much unaware of their weight.

PR #107, opened 2021-04-29
force Linux line endings also on Windows

mimmi20:
Signed-off-by: Thomas Müller <mimmi20@live.de>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

As Windows user I get all line endings changed to "\r\n" during checkout, and the coding style check fails. If I change the line endings, I see all files as changed. To make this easier and prevent any unwanted changes in the codebase, I prefer to force linux line endings with settings in the gitattributes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,10 +1,25 @@
-/.coveralls.yml export-ignore
-/.gitattributes export-ignore
-/.github/ export-ignore
-/.gitignore export-ignore
-/.markdownlint.json export-ignore
-/docs/ export-ignore
-/mkdocs.yml export-ignore
-/phpcs.xml export-ignore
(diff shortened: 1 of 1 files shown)


froschdesign:
@mimmi20 
Please check your local Git configuration on this topic: "[Formatting and Whitespace](http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace)"

mimmi20:
If I understand that text correct, I may only may tell git how to handle line ending changes if I want to commit a change, but I cannot say "dont convert LF to CRLF, but change CRLF to LF".

PR #108, opened 2021-04-29
render boolean attributes without value for HTML5

mimmi20:
Signed-off-by: Thomas Müller <mimmi20@live.de>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes/no (if you see the chaged rendering result as BC then "yes", "no" otherwise)
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

HTML5 allows to write boolean attributes like `disabled` or `required` to write without their values, if they are true. As a result the attribute will not be rendered as `disabled="disabled"`, but only as `disabled`.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1664,12 +1664,13 @@
       <code>$this-&gt;booleanAttributes[$attribute]['on']</code>
       <code>$this-&gt;booleanAttributes[$key]['off']</code>
     </MixedArrayAccess>
-    <MixedAssignment occurrences="10">
+    <MixedAssignment occurrences="11">
       <code>$attributes[$attribute]</code>
       <code>$escapedAttribute</code>
       <code>$prefix</code>
       <code>$prefix</code>
       <code>$prefix</code>
+      <code>$strings[]</code>
       <code>$this-&gt;doctypeHelper</code>
       <code>$this-&gt;escapeHtmlAttrHelper</code>
       <code>$this-&gt;escapeHtmlHelper</code>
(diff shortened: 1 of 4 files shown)


Slamdunk:
@froschdesign to me this doesn't seem a BC Break, but rather a bugfix to our library not adhering to HTML5 standards, WDYT?

froschdesign:
@Slamdunk 
> but rather a bugfix to our library not adhering to HTML5 standards

There are four different ways to specify an attribute in HTML5:

### Empty Attribute Syntax

```html
<input disabled>
```

### Unquoted Attribute Value Syntax

```html
<input disabled=disabled>
```

### Single-Quoted Attribute Value Syntax

```html
<input disabled='disabled'>
```

### Double-Quoted Attribute Value Syntax 

```html
<input disabled="disabled">
```

laminas-form uses the last one, so nothing wrong so far.

> to me this doesn't seem a BC Break

The HTML output changes with this extension and this can result in errors in tests. I think of systems like Cypress, laminas-test, Blackfire.io Player, etc.
The changes are available immediately after an update, without having to activate or change anything.

Slamdunk:
> There are four different ways to specify an attribute in HTML5:

Right, sorry: https://html.spec.whatwg.org/#boolean-attribute

So this must go into `3.0`

froschdesign:
@Slamdunk 
Please don't get me wrong, it would be nice if we could release this feature as soon as possible, but I just want to avoid possible trouble or frustration for the user.

([PHPUnit DOM Assertions](https://github.com/lstrojny/phpunit-dom-assertions) is another example which you know well. 😉 )

Slamdunk:
Sorry I don't get what you want to communicate to me.

BTW, `3.0` needs to be shipped with https://github.com/laminas/laminas-form/pull/104 so no `3.0` will be tagged soon ^^

mimmi20:
Is it possible to avoid the BC break by adding an option for this? Would it be allowed to distribute this option from the form ViewHelper down to the elements?

Slamdunk:
Thank you @mimmi20

PR #110, opened 2021-05-13
Mark annotation manager and parser as deprecated for 2.17.0 release

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Since the annotation manager and parser from `laminas/laminas-code` have been removed in `4.0`, `laminas/laminas-form` should switch to `doctrine/annotation-parser` as discussed in #104. This MR marks some elements as deprecated, which will need to be dropped in the next major release, for switching to `doctrine/annotations`.

Additionally, this MR enables annotation parsing as part of the unit tests, which has been disabled for some unknown reasons.

**Milestone:** 2.17.0

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -27,7 +27,7 @@
     <env name="TESTS_LAMINAS_OB_ENABLED" value="false"/>
     <!-- Enable this if you have installed Doctrine\Common on the
              include_path or via composer. -->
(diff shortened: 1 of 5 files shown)


Slamdunk:
Thank you

PR #111, opened 2021-05-13
Added support for annotations using doctrine/annotations and removed deprecated code

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes 
| RFC           | yes
| QA            | no

### Description

As discussed in #101, the current conflicts constraint regarding `laminas-code` brings some troubles, especially because earlier `2.16.x` versions were released without that constraint (though not different in functionality). This leads to composer downgrading from `2.16.3` to earlier versions, which then will still trigger the errors mentioned in #100, as `2.16.0`, `2.16.1` and `2.16.2` can be installed together with `laminas-code:4.0`, which does not have support for annotations anymore.

Therefore, to keep support for annotations, a new solution has to be found. Since `laminas-code:3.x` was relying on `doctrine/annotations`, for me it makes sense to directly implement that functionality in `laminas-form`. This is what this PR provides. 

This MR is a follow-up to #104 and needs a new major release due to BC breaks. To make upgrading more or less convenient for users, I have created MR #110 which is closely related to this MR as it marks all elements as deprecated for a 2.17.0 release, which this MR drops.

Additionally, as this MR aims at a 3.0 release, I have removed all other elements that have been marked as deprecated in earlier releases.

**Milestone:** 3.0.0


driehle:
@Slamdunk Sorry for creating a new MR on this, but this one now is targeted at a 3.0 release. There are some further changes which I'd suggest for a 3.0 release:

 - dropping `laminas/laminas-zendframework-bridge` and remove `replace` section in `composer.json`
 - dropping `laminas/laminas-cache` from `require-dev` (looks like its not used anywhere?!)
 - refactor all factories to `laminas-servicemanager:^3.0` and add a conflict for `laminas-servicemanager:<3.0`
 - mark `ElementAnnotationsListener` and `FormAnnotationsListener` as final and internal
 - mark all factories as final
 - mark `AnnotationBuilder`, `AnnotationCollection` and all annotation classes as final

Let me know what you thing and if I should add this in this MR here or on a separate one.

docs/book/quick-start.md
```diff
@@ -956,8 +956,6 @@ to define all of these?
 Annotations allow us to solve this problem. You can define the following
 behaviors with the shipped annotations in laminas-form:
 
-- `AllowEmpty`: mark an input as allowing an empty value. This annotation does
-  not require a value.
```
froschdesign: This feature is still present in laminas-inputfilter and there are also no plans to change this in the _near_ future.

docs/book/quick-start.md
```diff
@@ -1007,9 +1003,6 @@ behaviors with the shipped annotations in laminas-form:
   of validator options for the constructor:
   `@Validator({"name": "StringLength", "options": {"min":3, "max": 25}})`.
   This annotation may be specified multiple times.
-- `ContinueIfEmpty`: indicate whether the element can be submitted when it
-  is empty. A boolean value is expected. If `@Required` is set to `false`, this
-  needs to be set to `true` to allow the field to be empty.
```
froschdesign: The same here, the feature is deprecated in laminas-inputfilter but at the moment there is no plan for a new major version.

docs/book/quick-start.md
driehle: @froschdesign Good point, I'll restore `AllowEmpty` and `ContinueIfEmpty`.

Slamdunk:
Thank you very much @driehle for your work. As soon as I have the focus and time, I'll review this meticulously so we can finally drop `laminas/code`

Slamdunk:
This is good to me to go, waiting for other reviewers

Slamdunk (approved):

src/Annotation/AnnotationCollection.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+/**
+ * @see       https://github.com/laminas/laminas-code for the canonical source repository
```
boesing: license headers are not allowed anymore. 
using codestandard 2.2+ will verify that. 

would be good if you take cover on that manually in this PR.

src/Annotation/AnnotationCollection.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Laminas\Form\Annotation;
+
+use ArrayObject;
+
+use function get_class;
+
+/**
+ * @extends ArrayObject<array-key, object>
+ */
+class AnnotationCollection extends ArrayObject
```
boesing: can we mark this final? imho no need to provide abstraction support here

Slamdunk: Sure, done, thanks

src/Annotation/AnnotationCollection.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+final class AnnotationCollection extends ArrayObject
+{
+    /**
+     * Checks if the collection has annotations for a class
+     *
+     * @param  string $class
+     * @return bool
+     */
+    public function hasAnnotation($class)
```
froschdesign: Type declarations and return types should be specified for all new methods.

Slamdunk (approved):

src/Annotation/ElementAnnotationsListener.php
```diff
@@ -315,18 +317,16 @@ public function handleInputAnnotation($e)
...
-     * Handle the Object and Instance annotations
+     * Handle the Instance annotations
      *
      * Sets the object to bind to the form or fieldset
      *
      * @param  EventInterface $e
      * @return void
      */
-    public function handleObjectAnnotation($e)
+    public function handleInstanceAnnotation($e)
```
froschdesign: Type declaration and return type is missing, DocBlock can also be reduced.

src/Annotation/FormAnnotationsListener.php
```diff
@@ -123,18 +123,16 @@ public function handleInputFilterAnnotation($e)
...
-     * Handle the Object and Instance annotations
+     * Handle the Instance annotations
      *
      * Sets the object to bind to the form or fieldset
      *
      * @param  EventInterface $e
      * @return void
      */
-    public function handleObjectAnnotation($e)
+    public function handleInstanceAnnotation($e)
```
froschdesign: The same here.

froschdesign (changes requested):

src/Annotation/ElementAnnotationsListener.php
Slamdunk: They are in place as DocBlock, as you see.
I'd like to have a massive cleanup for `3.0.0` for all these things, including these ones.

Is it ok for you to have this change in a separate PR so we have smaller diff here?

src/Annotation/AnnotationCollection.php
driehle: @froschdesign should we, given that this goes in 3.0 anyways, add return types and type declarations to all methods?

Slamdunk: Indeed, that's what I want too: https://github.com/laminas/laminas-form/pull/111#discussion_r637832963

froschdesign: If the methods are new, why wait?

Slamdunk: Done

boesing (approved):

src/Annotation/ElementAnnotationsListener.php
driehle: @Slamdunk If desired, I can provide a PR for such things once the open MRs have been merged.

Slamdunk:
Thank you @driehle , I'll take case of the rebase for you other MR

PR #112, opened 2021-05-14
Dropped support for servicemanager 2.x and removed unused dependencies

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes (unused dependencies)
| BC Break      | yes (dropped support for servicemanager 2.x)
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

This MR is based on top of #111 and adds:

 - dropping `laminas/laminas-zendframework-bridge` and remove `replace` section in `composer.json`
 - dropping `laminas/laminas-cache` from `require-dev` (looks like its not used anywhere?!)
 - refactor all factories to `laminas-servicemanager:^3.0` and add a conflict for `laminas-servicemanager:<3.0`

**Milestone:** 3.0.0

PR #113, opened 2021-05-16
Added support for PHP 8 attributes as annotations alternative

### LINKED ISSUE #48: Form annotations don't work on PHP 7
I've narrowed the problem down the the Zend-code dependency. I don't know if there is any way we can fix this for PHP 7; perhaps a documentation update warning of this fact. ~~The fix I assume is going to be to upgrade to Zend-code 3 which has PHP 7 support.~~ Turns out even Zend code 3.0.0 isn't sufficient

---

Originally posted by @carnage at https://github.com/zendframework/zend-form/issues/51

### LINKED ISSUE #84: [RFC: PHP 8 Support] Replace Form Annotations with PHP 8 Attributes
As of current ongoing PHP8-discussions and Pull Requests regarding support for Form Annotations I suggest:
How about using the natively supported introduced PHP Attributes https://www.php.net/manual/de/language.attributes.overview.php from PHP 8 instead?

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes
| RFC           | yes
| QA            | no

### Description

This MR is based on top of #111 and #112 and adds support for using PHP 8-style attributes. Those can be used as an alternative to PHPDoc annocations, which are still supported.

Please note that all annotation classes have been refactored in order to be compatibel with PHP 8 attributes. However, the usage of these annotations (in DocBlock mode) remains the same. So even though this is a BC break, users will not need to update their annotations unless they actively want to migrate to PHP 8 attributes.

Examples can be found in the `tests/TestAsset/Annotation/` folder, e.g.:

```php
/**
 * @Annotation\Name("some_name")
 * @Annotation\Attributes({"legend":"Some Fieldset"})
 * @Annotation\InputFilter("LaminasTest\Form\TestAsset\Annotation\InputFilter")
 * @Annotation\ValidationGroup({"omit", "keep"})
 */
#[Annotation\Name("some_name")]
#[Annotation\Attributes(["legend" => "Some Fieldset"])]
#[Annotation\InputFilter("LaminasTest\Form\TestAsset\Annotation\InputFilter")]
#[Annotation\ValidationGroup(["omit", "keep"])]
class ClassEntity
{
    /**
     * @Annotation\Exclude()
     */
    #[Annotation\Exclude]
    public $omit;

    /**
     * @Annotation\Name("keeper")
     * @Annotation\Attributes({"type":"text"})
     */
    #[Annotation\Name("keeper")]
    #[Annotation\Attributes(["type" => "text"])]
    public $keep;
}
```

<del>As you can see, you can assign both DocBlock annotations and PHP 8 attributes to a class, though in reality you would obviously only use one of them.  If desired, I could add functions like `createFormFromAnnotations`, `createFormFromAttributes`,  `getFormSpecificationFromAnnotations` and `getFormSpecificationFromAttributes`. That would make the boolean flag obsolete. Alternatively, we could also add two constances like `AnnotationBuilder::USE_DOCBLOCK` and `AnnotationBuilder::USE_ATTRIBUTES`.</del>

As discussed, an `AttributeBuilder` has been introduced besides the existing `AnnotationBuilder`. Usage is as follows:

```php
$entity  = new TestAsset\Annotation\ClassEntity();
$annotationBuilder = new Annotation\AnnotationBuilder();
$attributeBuilder = new Annotation\AttributeBuilder();

// use DocBlock annotations -> nothing changed, works for PHP 7.3, 7.4, 8.0!
$form = $annotationBuilder->createForm($entity);

// use PHP attributes, will obviously require PHP 8.0
$form = $attributeBuilder->createForm($entity);
```

Let me know what you think. :-)

**Milestone:** 3.0.0


driehle:
@boesing Issue was that `doctrine/annotations` version constraint was set too low, we'll need `^1.12.0` instead of `^1.11.0`. Pushed an update.

src/Annotation/AnnotationBuilder.php
```diff
@@ -164,24 +194,52 @@ public function getFormSpecification($entity)
...
 
     /**
      * Create a form from an object.
      *
      * @param  string|object $entity
+     * @param  bool $useAttributes true if PHP8 attributes should be parsed instead of PHPDoc annotations
      * @return FormInterface
      */
-    public function createForm($entity)
+    public function createForm($entity, bool $useAttributes = false)
```
Slamdunk: I am unsure about this change.

As a user I image I would just refactor everything from DocBlocks to Attributes and expect everything to work seamlessly. Having to set this new flag sounds weird to me.

I think we should move this flag into the constructor, the user then has to change only the related factory but not its API calls.
An update to `\Laminas\Form\Annotation\AnnotationBuilderFactory` that gathers the flag from the config would close the circle easily.

The approach I'm suggesting would be sort of an all-or-nothing, but I'm not comfortable changing so many APIs for this feature.

WDYT @boesing @froschdesign ?

src/Annotation/AllowEmpty.php
```diff
@@ -13,28 +14,24 @@
  * \Laminas\InputFilter\Input should enable the allowEmpty flag.
  *
  * @Annotation
+ * @NamedArgumentConstructor
```
Slamdunk, line 19: Import is missing

Slamdunk (changes requested):
Beware that I've rebased the branch in your repo, so check it locally as a new name.

src/Annotation/AnnotationBuilder.php
boesing: tbh, without rector which converts annotations to attributes, I would not release any Version which only supports attributes. 
Imho, the annotation builder should be used for annotation based forms and another attribute builder could focus on attributes. 
having the same service doing both does not fit into the single responsibility principle. 

maybe introducing a new builder which just handles some kind of "config" would be better than having a parser + builder in the same service.

----

As a user, I'd prefer not having to full rewriting all annotations by myself. I use plenty of annotations in a project and having to rewrite everything would be a pain I dont expect from an update.

Thats actually something where laminas was always good at. We should spend time in creating migration tooling or continue supporting annotations.

src/Annotation/AllowEmpty.php
boesing: afaik, this annotation was marked deprecated and thus should be dropped anyways? 

afair using the NotEmpty validator was the way to go.

src/Annotation/AnnotationBuilder.php
driehle: Initially, I had the same idea as well that building an `AttributeBuilder` parallel to the existring `AnnotationBuilder` might be a good idea. However, then we would probably also need to rename the namespace `\Laminas\Form\Annotation` to `\Laminas\Form\Attribute` and I am not sure if we want to do that. Sure, we could still add the attributes as an aliased/extended class in the old namespace for compatibility reasons and to ensure a smooth upgrade process.

Besides, I fully agree with @boesing that we should not drop annotation support at least until we drop support for PHP 7.4.

src/Annotation/AllowEmpty.php
driehle: See froschdesign's comment [here](https://github.com/laminas/laminas-form/pull/111#discussion_r632311433).

Slamdunk: Still the `use` statement for `@NamedArgumentConstructor` is missing from the other dozen Annotations, so please add it everywhere

driehle: done

src/Annotation/AnnotationBuilder.php
Slamdunk: I have no intention to drop nor deprecate annotations.

Let's try the two builder path: move the current `AnnotationBuilder` to an `AbstractBuilder` with an `abstract public function getFormSpecification($entity)` and implement the corresponding `getFormSpecificationFromAttributes` and `getFormSpecificationFromAnnotations` into two new classes: `AttributeBuilder` and `AnnotationBuilder`.
The check for PHP8 and Attributes can be moved from `getFormSpecification` to `AttributeBuilder::__constructor()`.

Regarding the whole `\Laminas\Form\Annotation` namespace: I think there is no need to rename it, the word `annotation` will stay for a long time to indicate these concepts.

src/Annotation/AbstractBuilder.php
```diff
@@ -0,0 +1,353 @@
...
+    /**
+     * Implementation of deriving a form specification from an entity
+     *
+     * Must return an array with two elements, where the first element is the form specification and the
+     * second element is the input filter specification.
+     *
+     * @param string|object $entity
+     * @return array
+     */
+    abstract public function getFormSpecificationInternal($entity): array;
```
Slamdunk:
```suggestion
    abstract protected function getFormSpecificationInternal($entity): array;
```
No need to expose this

src/Annotation/AnnotationBuilder.php
```diff
@@ -65,96 +23,19 @@ public function __construct()
...
-     * Get event manager
+     * Derive a form specification from doctrine annotations for a given entity
      *
-     * @return EventManagerInterface
+     * @param  object|string $entity
+     * @throws \ReflectionException
+     * @return array
      */
-    public function getEventManager()
+    public function getFormSpecificationInternal($entity): array
```
Slamdunk:
```suggestion
    protected function getFormSpecificationInternal($entity): array
```

src/Annotation/AttributeBuilder.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Derive a form specification from PHP attributes for a given entity
+     *
+     * @param  object|string $entity
+     * @throws \ReflectionException
+     * @return array
+     */
+    public function getFormSpecificationInternal($entity): array
```
Slamdunk:
```suggestion
    protected function getFormSpecificationInternal($entity): array
```

Slamdunk (changes requested):

src/Annotation/AbstractBuilder.php
driehle: Indeed, fixed.

src/Annotation/AnnotationBuilder.php
driehle: done

driehle:
@Slamdunk I am unsure about the Psalm error, could you help me out there?

src/Annotation/ContinueIfEmpty.php
```diff
@@ -13,26 +14,24 @@
  * \Laminas\InputFilter\Input should enable the continueIfEmpty flag.
  *
  * @Annotation
+ * @NamedArgumentConstructor
```
Slamdunk, line 19: Still missing the `use` statement

Slamdunk (changes requested):
No need to worry about PSalm, I'll take care of it.
Just few other missing imports and we are done for me

src/Annotation/ContinueIfEmpty.php
driehle: done

driehle:
Issue #84 and probably #48 should be linked

Slamdunk (approved):

src/Annotation/AbstractBuilder.php
```diff
@@ -0,0 +1,353 @@
...
+     */
+    protected $preserveDefinedOrder = false;
+
+    /**
+     * Set form factory to use when building form from annotations
+     *
+     * @param  Factory $formFactory
+     * @return $this
+     */
+    public function setFormFactory(Factory $formFactory)
```
froschdesign, line 56: @Slamdunk 
What is your current plan? Do you want to release another major version to add the return types or is this also a topic for version 3?

Slamdunk: I'm sorry I haven't documented it because I didn't expect @driehle doing such great PRs and postponing my job about types ^^
It is my intention for `v3` to have `laminas-codingstandard:v2` and reboot PSalm to have empty baseline and incrementally raise its level from 8 to whatever I can reach, and in the path I'll add types (or move them from DocBlock to code) everywhere.

froschdesign: > I'm sorry I haven't documented…

My suggestion: You can create a project here in the repository. There you can add notes for all your ideas and also issue reports and pull requests.

> …@driehle doing such great PRs…

Right! 👍 ❤️

driehle: The reason I am pushing this currently is that some libraries have already migrated to laminas-code v4 and their updates are blocked because of laminas-form with v2 still sticking to laminas-code v3. :angel:

Slamdunk:
Thank you @driehle

PR #116, opened 2021-05-25
Element\Select selecting incorrect values when using floats (issue #52)

### LINKED ISSUE #52: When calling setValue Element\Select, selected incorrect values
When calling $select->setValue(array('1.1')), selected values French and Japanese,
but Japanese is incorrect value

``` php
use Zend\Form\Element;
use Zend\Form\Form;

$select = new Element\Select('language');
$select->setLabel('Which is your mother tongue?');
$select->setValueOptions(array(
    '1.1' => 'French',
    '1.2' => 'English',
    '1.10' => 'Japanese',
    '1.20' => 'Chinese',
));

$form = new Form('language');
$form->add($select);
```

---

Originally posted by @googlle at https://github.com/zendframework/zend-form/issues/18

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This MR fixes #52 (and zendframework/zend-form#18) by adding the test case provided in #52 and an implementation to let that test pass.

I would consider this a BC break since as people who might have relied on the currently implemented weak comparison in v2.x might be affected by this change. However, that will probably be an edge case.

### DIFF
--- a/src/View/Helper/FormSelect.php
+++ b/src/View/Helper/FormSelect.php
@@ -6,8 +6,8 @@ use Laminas\Form\Element\Hidden;
 use Laminas\Form\Element\Select as SelectElement;
 use Laminas\Form\ElementInterface;
 use Laminas\Form\Exception;
-
 use Laminas\Stdlib\ArrayUtils;
+
 use function array_key_exists;
 use function array_merge;
 use function compact;
--- a/src/View/Helper/FormSelect.php
+++ b/src/View/Helper/FormSelect.php
@@ -7,6 +7,7 @@ use Laminas\Form\Element\Select as SelectElement;
 use Laminas\Form\ElementInterface;
 use Laminas\Form\Exception;
 
+use Laminas\Stdlib\ArrayUtils;
 use function array_key_exists;
 use function array_merge;
 use function compact;
@@ -204,7 +205,7 @@ class FormSelect extends AbstractHelper
                 $disabled = $optionSpec['disabled'];
             }
 
-            if ($this->inArrayWithStringConversion($value, $selectedOptions)) {
+            if (ArrayUtils::inArray($value, $selectedOptions, true)) {
                 $selected = true;
             }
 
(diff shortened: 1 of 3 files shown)


froschdesign:
@driehle 
It looks like there is too much in the branch. 😉

driehle:
> It looks like there is too much in the branch. 😉

I purposely based it on #113 since that one is already in review. Once that is merged only one commit will remain here. :-)

Slamdunk:
@driehle please rebase this PR

driehle:
done

src/View/Helper/FormSelect.php
```diff
@@ -205,7 +204,7 @@ public function renderOptions(array $options, array $selectedOptions = [])
                 $disabled = $optionSpec['disabled'];
             }
 
-            if (ArrayUtils::inArray($value, $selectedOptions)) {
+            if ($this->inArrayWithStringConversion($value, $selectedOptions)) {
```
Slamdunk: There is no need for a separate method, just use the `strict` variant: `ArrayUtils::inArray($value, $selectedOptions, true)`

Slamdunk (changes requested):

src/View/Helper/FormSelect.php
driehle: Unfortunatly, that won't work. Strict comparison failes because the unit tests assume that objects are identical when their string representation equals, but enabling strict comparison requires objects to be identical. Hence, if I replace the line as you suggested, the following unit tests fail:

```
1) LaminasTest\Form\View\Helper\FormSelectTest::testCanMarkOptionsAsSelectedWhenEmptyOptionOrZeroValueSelected
Failed asserting that '<select name="foo"><option value="">empty</option>\n
<option value="0">label0</option>\n
<option value="1">label1</option></select>' contains "<option value="0" selected="selected">label0</option>".

/home/driehle/PhpstormProjects/laminas-form/test/View/Helper/FormSelectTest.php:390

2) LaminasTest\Form\View\Helper\FormSelectTest::testRenderElementWithObjectIdentifiers
Failed asserting that '<select name="foo"><option value="42">This is the first label</option>\n
<option value="43">This is the second label</option></select>' matches PCRE pattern "#option .*?value="42" selected="selected"#".

/home/driehle/PhpstormProjects/laminas-form/test/View/Helper/FormSelectTest.php:441
```

To circumstance this, I have added the new method, which first converts objects to string and then performs a strict comparison. Such functionality is not provided in stdlib, I think.

Slamdunk: You are right, thank you

Issue #117, opened 2021-05-25
Reboot PSalm


Slamdunk:
PSalm was introduced in https://github.com/laminas/laminas-form/pull/99 without any following fix up.

We need to:

1. Remove the current baseline
1. Set PSalm error level to the lowest level
1. Fix everything fixable with as few baseline as possible
1. Release the result and enforce the reached level for new and ongoing PR
1. Repeat the process from [2] raising the level one step above

Slamdunk:
For now I won't go further than this :hot_face:

PR #118, opened 2021-05-25
Elements MonthSelect, DateSelect and DateTimeSelect differentiating between null and empty strings (issue 54)

### LINKED ISSUE #54: Calling SetValue on Element/DateSelect with an empty string set's to todays date

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | yes
| RFC           | yes
| QA            | no

### Description

This MR picks up the unit tests provided in MR #70 and fixes issue #54.

First, this MR prevents the PHP Notices `Trying to access array offset on value of type null` mentioned in #54, which should be considered a bugfix. 

Next, this MR aligns the behaviour of MonthSelect, DateSelect and DateTimeSelect. Previously, only DateTimeSelect had an explicit check for `null` values and handled such by using a current timestamp. Since DateTimeSelect has been aligned with the two other (that had a consistent behaviour anyways), the change für DateTimeSelect needs to be considered as a BC break.

Lastly, this MR provides a new feature for `DateTimeSelect`. The current behaviour was to return a crazy mixture of fields that have been set and fields that have not been set. For instance, when `createEmptyOption` was enabled, then `getValue()` could return something like `2021-05-25 ::`. Now, if the date is all null it will be replaced with the current date and if the time is all null it will become `00:00:00`.

Potentially, this could be evolved further for MonthSelect and DateSelect, so that `2021--` would become `2021-00-00`. The latter would at least syntactically what one would excpect to get from `getValue()`. This is how it could look like:

```php
    public function getValue()
    {
        $year = $this->getYearElement()->getValue();
        $month = $this->getMonthElement()->getValue();
        $day = $this->getDayElement()->getValue();

        if ($this->shouldCreateEmptyOption() && null === $year && null === $month && null === $day) {
            return null;
        }

        return sprintf(
            '%04d-%02d-%02d', 
            $year,
            $month,
            $day
        );
    }
```

### DIFF
--- a/src/Element/DateSelect.php
+++ b/src/Element/DateSelect.php
@@ -13,6 +13,7 @@ use Laminas\Validator\Date as DateValidator;
 use Laminas\Validator\ValidatorInterface;
 
 use function array_merge;
+use function is_array;
 use function is_string;
 use function sprintf;
 
@@ -107,7 +108,7 @@ class DateSelect extends MonthSelect
             }
         }
 
-        if (null === $value && !$this->shouldCreateEmptyOption()) {
+        if (null === $value && ! $this->shouldCreateEmptyOption()) {
             $value = new PhpDateTime();
         }
 
@@ -134,9 +135,9 @@ class DateSelect extends MonthSelect
 
     public function getValue(): string
     {
-        $year = $this->getYearElement()->getValue();
+        $year  = $this->getYearElement()->getValue();
         $month = $this->getMonthElement()->getValue();
-        $day = $this->getDayElement()->getValue();
+        $day   = $this->getDayElement()->getValue();
 
         if ($this->shouldCreateEmptyOption() && null === $year && null === $month && null === $day) {
             return null;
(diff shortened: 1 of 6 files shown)


driehle:
Rebased

Slamdunk:
I am unsure about this one though, I'd prefer to start #117 first

driehle:
Sure, no worries. I think for this one here we would probably need further input from @froschdesign and @boesing, especially on the proposed `sprintf('%04d'...)` change. However, to me such change would make a lot of sense as from my understanding the elements should either return timestamp-formatted string (e.g. `2021-00-00 00:00:00`) or null, but never something like `2021--01 00::48`, which currently would be possible.

Slamdunk:
Thank you @driehle

PR #119, opened 2021-05-26
Upgrade to laminas-coding-standard v2

Slamdunk:


driehle:
Commit 18927d6 has removed all `use Attribute;` statements from annotations, breaking unit tests with PHP 8.0.

Slamdunk:
Yup, I know, that's because PHP Code Sniffer doesn't see `#[Annotation]` as a valid use for an import. I'm trying to suppress the sniff for this case

Slamdunk:
Bug reported to Slevomat: https://github.com/slevomat/coding-standard/issues/1228

PR #121, opened 2021-05-26
PSalm: clear level 7 errors

### LINKED ISSUE #117: Reboot PSalm

Slamdunk:


psalm-baseline.xml
```diff
@@ -6,4 +6,9 @@
       <code>\Zend\Form\FormElementManager::class           =&gt; FormElementManager::class</code>
     </DuplicateArrayKey>
   </file>
+  <file src="src/FormElementManager.php">
+    <ParamNameMismatch occurrences="1">
+      <code>$name</code>
```
Slamdunk, line 11: This is because `\Laminas\ServiceManager\AbstractPluginManager::get` itself is inconsistent with `\Psr\Container\ContainerInterface::get` parameter names, so there's no way to fix this here until laminas-servicemanager fix it there.

PR #123, opened 2021-05-26
Update documentation regarding annotations

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

#122

Issue #71, opened 2020-07-30
A flexible possibility for InputErrorClasses (Ready for Bootstrap 4+)


mano87:
Hello,
currently it is only possible with great effort to use the forms in combination with Bootstrap 4+. It would be good if you could use the InputErrorClass independently from the helper formRow.

Or is there a better way to implement the attached example? See the following code snippets...

```php
/** @var Laminas\Form\View\Helper\FormElementErrors $helper */
$helper = $this->formElementErrors();
$helper->setMessageOpenFormat('<small class="invalid-feedback">');
$helper->setMessageCloseString('</small>');
$helper->setMessageSeparatorString('');
```

```php
<div class="form-group row">
    <?php echo $this->formLabel($matriculationnumber); ?>
    <div class="col-sm-10">

        <?php if ($matriculationnumber->getMessages()) {
            $classAttributes = $matriculationnumber->hasAttribute('class') ? $matriculationnumber->getAttribute('class') . ' ' : '';
            $classAttributes = $classAttributes . 'is-invalid';
            $matriculationnumber->setAttribute('class', $classAttributes);
        } ?>

        <?php echo $this->formElement($matriculationnumber); ?>
        <?php echo $this->formElementErrors()->render($matriculationnumber); ?>
    </div>
</div>
```

froschdesign:
Related to #62 and #63

driehle:
@mano87 Until laminas-form supports this, I can recommend you @neilime's [twbs-helper-module](https://github.com/neilime/twbs-helper-module), which extends the helpers, changes the HTML accordingly and overrides the service manager's config to transparently use the extended helpers. That way, you will have all markup rendered according to Boostrap 4+.

froschdesign:
> …Boostrap 4+.

At the moment, version 5 is not yet supported: https://github.com/neilime/twbs-helper-module/issues/165
But there is already a first pull request: https://github.com/neilime/twbs-helper-module/pull/167

froschdesign:
@mano87 
> ```php
> /** @var Laminas\Form\View\Helper\FormElementErrors $helper */
> $helper = $this->formElementErrors();
> $helper->setMessageOpenFormat('<small class="invalid-feedback">');
> $helper->setMessageCloseString('</small>');
> $helper->setMessageSeparatorString('');
> ```

[...]

Issue #62, opened 2020-04-01
FormRow: allow attributes for <fieldset> wrapper


Slamdunk:
|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

Hi, so far the `<fieldset>` wrapper for multi-checkbox, radio and so on can't have attributes since the rendered html is hard-coded:

https://github.com/laminas/laminas-form/blob/c3ad4db5a26a8c99a812463c2eea6244887bf18a/src/View/Helper/FormRow.php#L193-L202

For the multi-checkbox, an hypothetical `$multiCheckbox->setAttribute('fieldset_class' => 'foo');` could be retrieved and injected: we should think of a comprehensive solution for all the involved elements though.

froschdesign:
@Slamdunk 
Thanks for this request!

> For the multi-checkbox, an hypothetical `$multiCheckbox->setAttribute('fieldset_class' => 'foo');` could be retrieved and injected: we should think of a comprehensive solution for all the involved elements though.

The question which arise here: is the FormRow helper the correct place to handle the fieldset?

The helper FormCollection already contains a "wrapper" functionality. Maybe we can adopt this functionality to all helper with multiple elements and also for the FormRow helper itself. This would allow to create custom wrappers for the different element types and to create a markup with the FormRow helper which is used by CSS frameworks like Bootstrap, Tailwind and Co.

Please attach further ideas and thoughts!

Slamdunk:
> The helper FormCollection already contains a "wrapper" functionality. Maybe we can adopt this functionality to all helper with multiple elements and also for the FormRow helper itself.

That's indeed a good suggestion :+1:

> [...] to create a markup with the FormRow helper which is used by CSS frameworks like Bootstrap, Tailwind and Co

In fact that's what I was looking for: avoid custom styling and rely solely to external CSS tools. This should be the path for any view helper these days: classes allowed everywhere :ship:

froschdesign:
I think the goal for _standard_ forms should be:

```php
<?= $this->form($form) ?>
```

which means we need support for configurations like this:

```php
return [
    'view_helper_config' => [
        'form_row' => [
            'input_error_class' => 'is-invalid',
            'wrapper'           => '…',
            'label_wrapper'     => '…',
        ],
    ],
];
```

or _(pseudo code)_

```php
return [
    'view_helper_config' => Laminas\Form\View\Helper\HelperConfig::BOOTSTRAP,
];
return [
[...]

PR #128, opened 2021-05-27
Promote DocBlock to native types

Slamdunk:


src/Annotation/BuilderAbstractFactory.php
```diff
@@ -73,7 +71,7 @@ public function canCreate(ContainerInterface $container, $requestedName)
      *
      * @return array
```
Slamdunk, line 72: PHP Code Sniffer keeps `@return array` notation to let be more specific if the developer is willing to be so

froschdesign:
@Slamdunk 
This is a big step. I think it would be good if we could add something to the migration guide to help the user, for example using Reactor with some rules for rewriting custom classes.

What do you think?

Slamdunk:
Even though it seems big, not even e single LOC of https://docs.laminas.dev/laminas-form/ needs to be changed (AFAICT).

The only and most common extension point is the `Form`, which typically only gets `init()` and `getInputFilterSpecification()` implemented, none of which requires a change.

Users so advanced to write their own `Elements` or `ViewHelpers` just need to promote the types already in use, for which I can suggest the command I've used if you wish:
```console
$ php-cs-fixer fix --verbose --rules=phpdoc_to_param_type,phpdoc_to_return_type --allow-risky=yes
```

froschdesign:
@Slamdunk 
I mean, a big step in a good and right direction! 😃 

> Users so advanced to write their own `Elements` or `ViewHelpers` just need to promote the types already in use, for which I can suggest the command I've used if you wish:

That would be great! 👍

Slamdunk:
I don't see an UPGRADE file or page, is there a standard to write and publish it in `laminas` org?

driehle:
I created a `migration/v2-to-v3.md` in docs with #123, in accordance to other components, e.g. https://docs.laminas.dev/laminas-router/migration/v2-to-v3/.

froschdesign:
The documentation is the right place for this and a new page is on the way with #123 

https://github.com/laminas/laminas-form/blob/17622276cefbe3f32a64d7623c3d0bc68a8e5b58/docs/book/migration/v2-to-v3.md

docs/book/migration/v2-to-v3.md
```diff
@@ -4,15 +4,27 @@ laminas-form v3 makes a number of changes that may affect your application. This
...
+laminas-form v3 promoted `@param` and `@return` annotations of non-mixed types to
+the corresponding function signatures. This change should affect only highly customized
+element, fieldset or form classes.
+If you need to automate native types promotion, [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)
+can help you with this command:
+
+```console
+$ php-cs-fixer fix --rules=phpdoc_to_param_type,phpdoc_to_return_type --allow-risky=yes path/to/my/custom/forms/
+```
+
```
froschdesign, line 18: @Slamdunk 
Thank you! 😃

Slamdunk:
@driehle @froschdesign I'd like to release `3.0.0` after https://github.com/laminas/laminas-form/pull/98#issuecomment-850383139 

I've already tested `3.0.x-dev` in my app and found few bugs like #133, #135 and #136 linked to this PR: could you please test `3.0.x-dev` too in the upcoming days?

driehle:
I am wondering if we should probably consider #62 and #71 before the release of 3.0.0?

froschdesign:
@driehle 
There is still some time before we can release version 3.0.0, because I have to extend the documentation. The versioning is missing and some other updates.

Help is welcome at any time! 😃

Slamdunk:
@driehle since #62 and #71 can be addressed in a minor, this morning I've released `3.0.0` to speed the things up.

Thank you very much for the big help you gave us releasing this important milestone :+1:

PR #130, opened 2021-05-27
Allow name attribute for <fieldset>

### LINKED ISSUE #42: formCollection View Helper doesn't render name
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

---

Original Issue: https://api.github.com/repos/zendframework/zendframework/issues/7687
User: @gregGit
Created On: 2016-03-14T16:20:39Z
Updated At: 2016-04-13T07:48:07Z
Body
When using form view helper, the formCollection view helper is used to render fieldset. 
But form some reason that i don't understand this view helper doesn't set the name attribute in the fieldset (unset($attributes['name']); in the render Method).
I belive name is a valid html attributes for fieldset i don't understand why ZF2 never render this tag in a fieldset, and it can be usefull to have it in the DOM.

For test create an action like this :

```
  $fs = new \Zend\Form\Fieldset('my-fieldset');
        $fs->setAttribute('name', 'my-fieldset');
        $fs->add(array(
            'type' => 'text',
            'name' => 'text',
            'options' => array(
                'label' => 'The Text'
            )
        ));

        $fs->add(array(
            'type' => 'text',
            'name' => 'title',
            'options' => array(
                'label' => 'Blog Title'
            )
        ));


        $form=new \Zend\Form\Form('my-form');
        $form->add($fs);
        return new ViewModel(array('form' => $form));
```

and just " echo $this->form($form)" in the view. 

No attribute name for the fieldset is rendered.

---
### Comment

User: @froschdesign
Created On: 2016-04-13T07:48:07Z
Updated At: 2016-04-13T07:48:07Z
Body

> I belive name is a valid html attributes for fieldset

It's only allowed in HTML5.

---

---

Originally posted by @GeeH at https://github.com/zendframework/zend-form/issues/86

Slamdunk:
Close #42

### DIFF
--- a/src/View/Helper/FormCollection.php
+++ b/src/View/Helper/FormCollection.php
@@ -8,6 +8,7 @@ use Laminas\Form\Element\Collection as CollectionElement;
 use Laminas\Form\ElementInterface;
 use Laminas\Form\FieldsetInterface;
 use Laminas\Form\LabelAwareInterface;
+use Laminas\View\Helper\Doctype;
 use Laminas\View\Helper\HelperInterface;
 use RuntimeException;
 
@@ -18,6 +19,15 @@ use function sprintf;
 
 class FormCollection extends AbstractHelper
 {
+    /**
+     * Attributes valid for this tag (form)
+     *
+     * @var array
+     */
+    protected $validTagAttributes = [
+        'name' => true,
+    ];
+
     /**
      * If set to true, collections are automatically wrapped around a fieldset
      *
@@ -66,6 +76,12 @@ class FormCollection extends AbstractHelper
      * @var HelperInterface
      */
     protected $fieldsetHelper;
+
+    /** @var array */
+    private $doctypesAllowedToHaveNameAttribute = [
+        Doctype::HTML5  => true,
+        Doctype::XHTML5 => true,
+    ];
 
     /**
      * Invoke helper as function
@@ -118,8 +134,10 @@ class FormCollection extends AbstractHelper
         // Every collection is wrapped by a fieldset if needed
         if ($this->shouldWrap) {
             $attributes = $element->getAttributes();
-            unset($attributes['name']);
-            $attributesString = $attributes ? ' ' . $this->createAttributesString($attributes) : '';
+            if (! isset($this->doctypesAllowedToHaveNameAttribute[$this->getDoctype()])) {
+                unset($attributes['name']);
+            }
+            $attributesString = $attributes !== [] ? ' ' . $this->createAttributesString($attributes) : '';
 
             $label  = $element->getLabel();
             $legend = '';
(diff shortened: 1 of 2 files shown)


src/View/Helper/FormCollection.php
```diff
@@ -118,8 +128,10 @@ public function render(ElementInterface $element): string
         // Every collection is wrapped by a fieldset if needed
         if ($this->shouldWrap) {
             $attributes = $element->getAttributes();
-            unset($attributes['name']);
-            $attributesString = $attributes ? ' ' . $this->createAttributesString($attributes) : '';
+            if ($this->getDoctype() === Doctype::XHTML1_STRICT) {
```
froschdesign: In HTML4 there is also no attribute `name` for the `fieldset` element.

Slamdunk: Indeed, I looked up HTML4 strict specs and you're right.
Are you able to tell me which of the supported Doctypes don't have it?

https://github.com/laminas/laminas-view/blob/c34b15f42cac410f35010163de169ab71c39bf73/src/Helper/Doctype.php#L22-L35

```php
class Doctype extends AbstractHelper
{
    /**#@+
     * DocType constants
     */
    const XHTML11             = 'XHTML11';
    const XHTML1_STRICT       = 'XHTML1_STRICT';
    const XHTML1_TRANSITIONAL = 'XHTML1_TRANSITIONAL';
    const XHTML1_FRAMESET     = 'XHTML1_FRAMESET';
    const XHTML1_RDFA         = 'XHTML1_RDFA';
    const XHTML1_RDFA11       = 'XHTML1_RDFA11';
    const XHTML_BASIC1        = 'XHTML_BASIC1';
    const XHTML5              = 'XHTML5';
    const HTML4_STRICT        = 'HTML4_STRICT';
    const HTML4_LOOSE         = 'HTML4_LOOSE';
    const HTML4_FRAMESET      = 'HTML4_FRAMESET';
    const HTML5               = 'HTML5';
    const CUSTOM_XHTML        = 'CUSTOM_XHTML';
    const CUSTOM              = 'CUSTOM';
}
```

froschdesign: The `name` attribute is only allowed for HTML5 and XHTML5. The attribute was introduced with HTML5.

test/View/Helper/FormCollectionTest.php
```diff
@@ -419,4 +420,45 @@ public function testForElementHelperNotInstanceOfHelperInterface(): void
...
+            [Doctype::XHTML1_RDFA,         false],
+            [Doctype::XHTML1_RDFA11,       false],
+            [Doctype::XHTML_BASIC1,        false],
+            [Doctype::XHTML5,              true],
+            [Doctype::HTML4_STRICT,        false],
+            [Doctype::HTML4_LOOSE,         false],
+            [Doctype::HTML4_FRAMESET,      false],
+            [Doctype::HTML5,               true],
+            [Doctype::CUSTOM_XHTML,        false],
+            [Doctype::CUSTOM,              false],
```
Slamdunk, line 461: @froschdesign is this `dataProvider` correct?

froschdesign (approved):
🚢  it!

PR #134, opened 2021-05-28
Adds factory for FormElementErrors view helper

froschdesign:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### DIFF
--- a/docs/book/helper/form-element-errors.md
+++ b/docs/book/helper/form-element-errors.md
@@ -72,3 +72,24 @@ Method signature                                                     | Descripti
 `getAttributes() : array`                                            | Returns the attributes that will go on the message open format.
 `setTranslateMessages(bool $flag) : self`                            | Indicate whether or not element validation error messages should be translated during `render()`. Default is to translate them.
 `render(ElementInterface $element [, array $attributes = array()]) : string` | Renders validation errors for the provided `$element`. Attributes provided will be used in the `messageOpenFormat`, and merged with any provided previously via `setAttributes()`.
+
+## Usage in a laminas-mvc Application
+
+Within a [laminas-mvc](https://docs.laminas.dev/laminas-mvc/) application, the options for the view helper can be set via the application configuration.
+
+Add the following lines to a [configuration file](https://docs.laminas.dev/laminas-mvc/services/#default-configuration-options), e.g. `config/autoload/global.config.php`:
+
+```php
+return [
+    'view_helper_config' => [
+        'form_element_errors' => [
+            'attributes'               => ['class' => 'help-inline'],
+            'message_close_string'     => '</div>',
+            'message_open_format'      => '<div%s>',
+            'message_separator_string' => '',
+            'translate_messages'       => false,
+        ],
+    ],
+    // …
+];
+```
--- a/src/View/Helper/Factory/FormElementErrorsFactory.php
+++ b/src/View/Helper/Factory/FormElementErrorsFactory.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Form\View\Helper\Factory;
+
+use Laminas\Form\View\Helper\FormElementErrors;
+use Psr\Container\ContainerInterface;
+
+use function array_key_exists;
+use function is_array;
+
+class FormElementErrorsFactory
+{
+    public function __invoke(ContainerInterface $container): FormElementErrors
+    {
+        $helper = new FormElementErrors();
+
+        if (! $container->has('config')) {
+            return $helper;
+        }
+
+        $config = $container->get('config');
+        if (isset($config['view_helper_config']['form_element_errors'])) {
+            $configHelper =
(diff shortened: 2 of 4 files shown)


docs/book/helper/form-element-errors.md
```diff
@@ -72,3 +72,24 @@ Method signature                                                     | Descripti
 `getAttributes() : array`                                            | Returns the attributes that will go on the message open format.
 `setTranslateMessages(bool $flag) : self`                            | Indicate whether or not element validation error messages should be translated during `render()`. Default is to translate them.
 `render(ElementInterface $element [, array $attributes = array()]) : string` | Renders validation errors for the provided `$element`. Attributes provided will be used in the `messageOpenFormat`, and merged with any provided previously via `setAttributes()`.
+
+## Usage in a laminas-mvc Application
+
+Within a [laminas-mvc](https://docs.laminas.dev/laminas-mvc/) application, the options for the view helper can be set via the application configuration.
```
Slamdunk, line 78:
```suggestion
Within a [PSR-11 container](https://www.php-fig.org/psr/psr-11/) like the one used in [laminas-mvc](https://docs.laminas.dev/laminas-mvc/) application, the options for the view helper can be set via the container configuration.
```

Slamdunk (changes requested):

docs/book/helper/form-element-errors.md
froschdesign: The section describes the usage in a laminas-mvc application and there the question arises: What is a "container configuration" in a laminas-mvc based application?  If we follow the documentation of laminas-mvc, then this does not exist.

I agree with you that we also need descriptions for other application types, but at the moment we only have something for laminas-mvc.

Slamdunk: The page has no reference to `laminas-mvc` except in this new chapter.
As the new functionality has no relationship with `laminas-mvc`, I'd like not to couple to it in the documentation neither.

Maybe we can change the title to `Usage in a PSR-11 container` and then add an example that could be valid to both `laminas-mvc` and `mezzio`, wdyt?

src/View/Helper/Factory/FormElementErrorsFactory.php
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Laminas\Form\View\Helper\Factory;
+
+use Laminas\Form\View\Helper\FormElementErrors;
+use Psr\Container\ContainerInterface;
+
+use function array_key_exists;
+use function is_array;
+
+class FormElementErrorsFactory
```
Slamdunk, line 13: The other factories we have have signatures for `laminas-servicemanager`.
I see this can be a standalone class, but for sake of consistency what about implementing `\Laminas\ServiceManager\Factory\FactoryInterface` here too?

docs/book/helper/form-element-errors.md
froschdesign: I am deliberately avoiding adding examples of other applications at the moment because we have no description how laminas-form can be used in Mezzio or any other application then laminas-mvc.
Yes, it's annoying, but why add something for a view helper when everything else is missing? I think it leads to more frustration than it helps when the basis is missing.

Btw. we have a project for this topic: https://github.com/orgs/laminas/projects/3

src/View/Helper/Factory/FormElementErrorsFactory.php
froschdesign: Two reasons why I do not use it here:

* the features `$requestedName` and `$options` of the interface are not needed
* `Laminas\ServiceManager\Factory\FactoryInterface` still uses the interface of container-interop

docs/book/helper/form-element-errors.md
Slamdunk: I agree on your concern, and I trust your much wider documenting experience :+1: 

I hope one day we can only refer to `PSR`s in docs, as I think `laminas-mvc` has somehow blocked the modular usage of `laminas` ecosystem, not technically but instinctively, as every doc refers to it and it seems like a magic black box that do everything and making transition to anything else impossible.

froschdesign: I totally agree with you, because I see it myself as one of the biggest problems of the documentation that the descriptions for integration in applications are missing. And even for the own MVC system! :disappointed:

Therefore, we also need the headings and must not hide sections like this. See the navigation on the left for the page content: https://docs.laminas.dev/laminas-form/helper/form-element-errors/
With the new section there is then one more: "Usage in a laminas-mvc Application"

src/View/Helper/Factory/FormElementErrorsFactory.php
froschdesign: Ah, we can mark this class as final.

Slamdunk: Ok, I'll make most classes `final` soon in a separate PR

Slamdunk (approved):

Slamdunk:
Thank you @froschdesign

PR #156, opened 2021-11-25
Limit type of fieldset/element constructor

japaveh:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

In Laminas Form v3 the \Laminas\Form\Fieldset->get() function requires a (string), this means that fieldsets can only have a string typed name. In V2 it was possible to use just an integer so the upgrade to V3 broke the code, but just at the retrieval of the fietset. 

The constructor of the fieldset (the formelement) casts the name to a string using setName but this moves the error to the getter which the results in an error you want to do $fiedset->get(1)

In this PR is narrowed the allowed types in the fieldset and element constructor to a nullable string, but maybe is ```string $name = null``` also fine, but I kept the null|string limitation so passing null should also be possible

This might break code like the string constraint on the getter but now on an earlier phase

### DIFF
--- a/src/Element.php
+++ b/src/Element.php
@@ -46,7 +46,7 @@ class Element implements
     protected $hasValue = false;
 
     /**
-     * @param  null|string   $name    Optional name for the element
+     * @param  null|int|string $name Optional name for the element
      * @param  iterable $options Optional options for the element
      * @throws Exception\InvalidArgumentException
      */
--- a/src/Element.php
+++ b/src/Element.php
@@ -46,7 +46,7 @@ class Element implements
     protected $hasValue = false;
 
     /**
-     * @param  null|int|string $name Optional name for the element
+     * @param  null|string $name Optional name for the element
      * @param  iterable $options Optional options for the element
      * @throws Exception\InvalidArgumentException
      */
--- a/src/Element.php
+++ b/src/Element.php
@@ -46,11 +46,11 @@ class Element implements
     protected $hasValue = false;
 
     /**
-     * @param  null|int|string   $name    Optional name for the element
+     * @param  null|string   $name    Optional name for the element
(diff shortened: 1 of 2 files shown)


japaveh:
I will update the tests :)

Slamdunk:
Pretty unfortunate oversight of mine :disappointed: 

So, I won't merge this bcause that's a BC Break and `laminas-form:v4` is anything but near in time.

But your concerns are right; to address that, I'd prefer to throw an Exception in case the provided name is an `int` and suggest to typecast it preemptively to string, so we can publish a bugfix release pretty soon.

Can you have a try on the approach I described please?

japaveh:
> Pretty unfortunate oversight of mine 😞
> 
> So, I won't merge this bcause that's a BC Break and `laminas-form:v4` is anything but near in time.
> 
> But your concerns are right; to address that, I'd prefer to throw an Exception in case the provided name is an `int` and suggest to typecast it preemptively to string, so we can publish a bugfix release pretty soon.
> 
> Can you have a try on the approach I described please?

You mean a typecast on the getter? There is already a typecast on the constructor. Maybe a typecast on the $fielset-get() is sufficient?

Slamdunk:
So, there is no way we can handle this nicely in `v3` major release.

This PR just moves type checks from static to runtime, which doesn't make much sense and adds no value.

https://github.com/laminas/laminas-form/pull/159 simply removes the possibility to pass integers to contructors, but cannot be done in `v3` because either static or runtime it would be a BC Break.

So I'm gonna mark https://github.com/laminas/laminas-form/pull/159 for `v4` milestone and close this.

By the way thank you very much for getting this issue emerged :clap:

PR #168, opened 2022-02-21
Prepare for Renovate with reusable workflows

ghostwriter:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Adding Renovate to Laminas, Mezzio repositories. https://gist.github.com/weierophinney/49ca8ae2b916d898dd3d4162b26ff41b

- Configure minimum supported PHP version
- Update composer dependencies
- Update to reusable `Continuous Integration` workflow
- Update to reusable `Automatic Releases` workflow

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,9 +9,9 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
-        "laminas/laminas-hydrator": "^4.3.0",
+        "laminas/laminas-hydrator": "^4.3.1",
         "laminas/laminas-inputfilter": "^2.13.0",
-        "laminas/laminas-stdlib": "^3.6.1"
+        "laminas/laminas-stdlib": "^3.7.1"
     },
     "conflict": {
         "doctrine/annotations": "<1.12.0",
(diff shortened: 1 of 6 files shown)


Ocramius (changes requested):
Requires rolling back to a much smaller `composer.lock` change (just hash change)

src/View/Helper/FormMultiCheckbox.php
```diff
@@ -412,6 +413,7 @@ protected function getInputHelper(): FormInput
             return $this->inputHelper;
         }
 
+        assert($this->view !== null);
```
Ocramius: I think baselining these would be more appropriate, as @gsteel keeps working on cleaning up the abstract helpers 🤔

Ocramius (changes requested):

composer.json
```diff
@@ -58,6 +58,9 @@
         "allow-plugins": {
             "composer/package-versions-deprecated": true,
             "dealerdirect/phpcodesniffer-composer-installer": true
+        },
+        "platform": {
+            "php": "7.4.99"
```
Ocramius, line 63: Lock file needs to be regenerated anyway

Ocramius (changes requested):

Ocramius (approved):
Nice! Thanks for generating a `composer.lock` 👍

PR #170, opened 2022-03-10
Allow override of CSRF name

### LINKED ISSUE #169: CSRF element overwrites name option with element name
### Bug Report

You can specify a custom name for the CSRF validator to use via `csrf_options`:

```php
$security = new Element\Csrf('security', [
    'csrf_options' => [
        'timeout' => null,
        'name' => 'login_csrf',
    ],
]);
```

This name however will be overwritten with the element's name by this piece of code: https://github.com/laminas/laminas-form/blob/d854a5610889ff42d36076a46bb2422f37caa29b/src/Element/Csrf.php#L75-L77

So above element would store the token under the name "security" instead of the expected "login_csrf" in the session. This is counter-intuitive as you would expect the options to take precedence over the element name.

|    Q        |   A
|------------ | ------
| Version(s)  | 3.1.1/3.2.x

#### Possible resolution

Inverting the order of arguments to use the element's name as a default instead of an override could fix this behavior:

```php
$csrfOptions = array_merge(['name' => $this->getName()], $csrfOptions);
```

villermen:
Allows the `csrf_options.name` option passed to a `Csrf` element to override the element's own name for the validator, instead of discarding the option entirely.

Fixes #169.

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no¹
| New Feature   | no
| RFC           | no
| QA            | no

¹ Technically yes for people who had the name CSRF option specified before (previously a no-op).

### DIFF
--- a/test/Element/CsrfTest.php
+++ b/test/Element/CsrfTest.php
@@ -83,4 +83,18 @@ final class CsrfTest extends TestCase
         $this->assertEquals(777, $validator->getTimeOut());
         $this->assertEquals('MySalt', $validator->getSalt());
     }
+
+    public function testNameOverride(): void
+    {
+        $element = new CsrfElement('foo');
+        $element->setOptions([
+            'csrf_options' => [
+                'name' => 'bar',
+            ],
+        ]);
+        $validator = $element->getCsrfValidator();
+        $this->assertEquals('foo', $element->getName());
+        $this->assertEquals('bar', $validator->getName());
+        $this->assertEquals('Laminas_Validator_Csrf_salt_bar', $validator->getSessionName());
+    }
 }
--- a/src/Element/Csrf.php
+++ b/src/Element/Csrf.php
@@ -73,7 +73,7 @@ class Csrf extends Element implements InputProviderInterface, ElementPrepareAwar
     {
         if (null === $this->csrfValidator) {
             $csrfOptions = $this->getCsrfValidatorOptions();
-            $csrfOptions = array_merge($csrfOptions, ['name' => $this->getName()]);
+            $csrfOptions = array_merge(['name' => $this->getName()], $csrfOptions);
             $this->setCsrfValidator(new CsrfValidator($csrfOptions));
             assert(null !== $this->csrfValidator);
         }


Slamdunk:
Weird that in the first commit, with the test but without the fix, 2 build passed the tests :thinking: 

WDYT may have happened?

test/Element/CsrfTest.php
```diff
@@ -83,4 +83,18 @@ public function testSetOptionsTraversable(): void
...
     }
+
+    public function testNameOverride(): void
+    {
+        $element = new CsrfElement('foo');
+        $element->setOptions([
+            'csrf_options' => [
+                'name' => 'bar',
+            ],
+        ]);
```
froschdesign, line 94: Is there a reason why the name for the validator should be different from the name for the form element?

villermen: @froschdesign In my case I'm going around our application making our CSRF tokens unique from eachother, but I don't want to touch the views or other usages of the elements.

Most of the form elements are named "security" meaning they all share the same tokens. Having to look for every case of `->get('security')` and matching it to the correct form and renaming it is going to be a tedious and error-prone endeavor. Being able to change just the namespace the token is stored in like this would greatly help me out.

villermen: @froschdesign Any update on this?

driehle:
@villermen I agree that the current implementation has the "issue" (might as well be a feature for some) that the token is unique for all forms where the element is named identically. Usually I name the element `csrf`, so the token will be the same for all forms. However, this can already be solved with the current implementation by providing your own session container. Please see the following `Form` class which I use as a base class for all of my forms:

```php
<?php

declare(strict_types=1);

namespace Application\Utility\Form;

use Laminas\Form\Element\Csrf;
use Laminas\Session\Container;

class Form extends \Laminas\Form\Form
{
    public function addCsrfElement($timeout = 600): void
    {
        $parts = explode('\\', static::class);
        $name = 'csrf' . $parts[0] . (count($parts) > 1 ? $parts[array_key_last($parts)] : '');
        $session = 'CsrfToken_' . $parts[0] . (count($parts) > 1 ? '_' . $parts[array_key_last($parts)] : '');

        $this->add([
            'type' => Csrf::class,
            'name' => 'csrf',
            'options' => [
                'csrf_options' => [
                    'name' => $name,
                    'session' => new Container($session),
                    'timeout' => $timeout,
                ],
            ],
        ]);
    }
}
```

However, looking at your changes I see that my option `csrf_options.name` indeed never changed anything and it was solely `csrf_options.session` which did the trick.

villermen:
@froschdesign It just feels very wrong to me that specifying the CSRF validator's options via `csrf_options` leaves out the most important option. The `name` option is documented to be exactly what you're looking for after reading https://docs.laminas.dev/laminas-form/v3/element/csrf/#multiple-csrf-elements-must-be-uniquely-named and then seeing https://github.com/laminas/laminas-validator/blob/bdd503adc83d814a5c94e598ea0eb9fc7ca56339/src/Csrf.php#L53, whereas the session option is not documented at all.

If the name is discarded as a feature I'd prefer it if the option were removed/not parsed altogether so it can not give the impression of improved security where there is none.

driehle:
Yeah, if I remember correctly, I found the `session` option only by reverse engineering. I'd definetly prefer the `name` option as well, as that one is clearly mentioned on the docs for exactly this purpose. I vote for merging this PR.

froschdesign:
@Slamdunk 
I would release this as a new functionality. Do you agree with that?

Slamdunk (approved):

Slamdunk:
Go for it :muscle:

froschdesign (approved):

froschdesign:
@villermen 
Thank you for your time and this contribution! 👍

PR #175, opened 2022-07-06
Extend global attributes array

Koopzington:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

With further development of the HTML standard there have been introduced a number of new global attributes of which some are also applicable on form elements. To support these new attributes i have decided to extend the existing whitelist of global attributes that's being used in the rendering process and added the attributes listed on [MDN](https://github.com/mdn/content/blob/3eb31ca205b37b6f4b3ea8cd278d2f2e32a56e96/files/en-us/web/html/global_attributes/index.md) which were missing.
As suggested by @froschdesign on slack i've commented out the ones i perceived as not really useful in context of form elements.

The following attributes have been added:
- [autocapitalize](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize) - automatically cases typed content
- [enterkeyhint](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint) - relevant for virtual keyboard contexts
- [inputmode](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) - relevant for virtual keyboard contexts
- [is](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is) - support for HTML custom elements
- [itemid](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/itemid) - commented out since Microdata on input elements didn't exactly make sense to me
- [nonce ](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce) - commented out since afaik they're only relevant for `<script>` and `<style>` or similar tags
- [part](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/part) - for CSS styling
- [slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot) - for shadow DOM usage
- [translate](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate) - commented out since the description points to Text nodes inside elements


src/View/Helper/AbstractHelper.php
```diff
@@ -93,18 +93,25 @@ abstract class AbstractHelper extends BaseAbstractHelper
      */
     protected $validGlobalAttributes = [
         'accesskey'          => true,
+        'autocapitalize'     => true,
```
Ocramius: Should we at least document where these attributes come from? A link to their spec would be helpful 🤔

Ocramius (changes requested):

src/View/Helper/AbstractHelper.php
Koopzington: you mean inside the code, per attribute?
`'autocapitalize' => true, // https://html.spec.whatwg.org/multipage/interaction.html#attr-autocapitalize`

I don't know how prone the spec page is regarding breaking links, should they ever decide to move that blurp to another page.

Or a mention on https://docs.laminas.dev/laminas-form/v3/helper/abstract-helper/?

Slamdunk: > you mean inside the code, per attribute?
> `'autocapitalize' => true, // https://html.spec.whatwg.org/multipage/interaction.html#attr-autocapitalize`

IMHO yes, per attribute pointing to `whatwg.org` is ok and enough. I wouldn't care about link stability, we can't be perfect.

> Or a mention on https://docs.laminas.dev/laminas-form/v3/helper/abstract-helper/?

Listing each attribute would be cumbersome to me, but a mention that only some are supported with a call to look at `Laminas\Form\View\Helper\AbstractHelper::$validGlobalAttributes` to have the full list, that would be very welcome

Ocramius (approved):
LGTM, thanks for taking the time to expand on the references.

As for the CS failures, I think those are on our sude to adjust/fix (probably suppress)

Koopzington:
Looks like i was a bit too hasty trying to fix things myself.
Should we use the snapshotted links now or not? :)

Ocramius:
I don't mind - these are usually very stable regardless 👍

Slamdunk (approved):
Thank you. I'll try to fix the CI once back from vacation :beach_umbrella:

Slamdunk:
Thank you @Koopzington

PR #177, opened 2022-07-17
Remove `container-interop/container-interop` in favour of `psr/container`

### LINKED ISSUE #171: Removal of `container-interop/container-interop`
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | yes

#### Summary

This issue is created to keep track of the status of getting rid `container-interop` in the `laminas` components.

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

Closes https://github.com/laminas/laminas-form/issues/171

### DIFF
--- a/docs/book/v2/advanced.md
+++ b/docs/book/v2/advanced.md
@@ -266,8 +266,8 @@ This now requires a factory to inject the form instance:
 namespace Application\Controller;
 
 use Application\Form\MyForm;
+use Interop\Container\ContainerInterface;
 use Laminas\Form\FormElementManager;
-use Psr\Container\ContainerInterface;
 
 class IndexControllerFactory
 {
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.29.0@7ec5ffbd5f68ae03782d7fd33fff0c45a69f95b3">
+<files psalm-version="4.30.0@d0bc6e25d89f649e4f36a534f330f8bb4643dd69">
   <file src="src/Annotation/AbstractBuilder.php">
     <DocblockTypeContradiction occurrences="1">
       <code>! is_string($entity)</code>
@@ -67,26 +67,13 @@
       <code>$this</code>
     </LessSpecificReturnStatement>
   </file>
-  <file src="src/Element/AbstractDateTime.php">
-    <LessSpecificImplementedReturnType occurrences="1">
-      <code>array</code>
-    </LessSpecificImplementedReturnType>
-  </file>
   <file src="src/Element/Captcha.php">
     <DocblockTypeContradiction occurrences="1">
       <code>gettype($captcha)</code>
     </DocblockTypeContradiction>
-    <LessSpecificImplementedReturnType occurrences="1">
-      <code>array</code>
-    </LessSpecificImplementedReturnType>
     <RedundantConditionGivenDocblockType occurrences="1">
       <code>is_object($captcha)</code>
     </RedundantConditionGivenDocblockType>
-  </file>
-  <file src="src/Element/Checkbox.php">
-    <LessSpecificImplementedReturnType occurrences="1">
-      <code>array</code>
-    </LessSpecificImplementedReturnType>
   </file>
   <file src="src/Element/Collection.php">
     <DocblockTypeContradiction occurrences="4">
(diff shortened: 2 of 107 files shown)


composer.json
```diff
@@ -19,7 +19,7 @@
         "laminas/laminas-eventmanager": "<3.4.0",
         "laminas/laminas-i18n": "<2.12.0",
         "laminas/laminas-recaptcha": "<3.4.0",
-        "laminas/laminas-servicemanager": "<3.10.0",
+        "laminas/laminas-servicemanager": "<3.14.0",
```
Ocramius: To switch without a BC break, `laminas/laminas-servicemanager` must be in the `require` section

Ocramius (changes requested):

docs/book/v2/advanced.md
```diff
@@ -265,8 +265,8 @@ This now requires a factory to inject the form instance:
 ```php
 namespace Application\Controller;
 
-use Interop\Container\ContainerInterface;
 use Application\Form\MyForm;
+use Psr\Container\ContainerInterface;
```
froschdesign: The current pull request adds the support for the PSR container interface for version 3 and not version 2. Therefore this change is wrong.

docs/book/v2/advanced.md
```diff
@@ -394,9 +394,9 @@ To enable this, we'll create a factory for our `AlbumFieldset` as follows:
 namespace Application\Form;
 
 use Album\Model\AlbumTable;
-use Interop\Container\ContainerInterface;
 use Laminas\ServiceManager\FactoryInterface;
 use Laminas\ServiceManager\ServiceLocatorInterface;
+use Psr\Container\ContainerInterface;
```
froschdesign: The same here.

froschdesign (changes requested):

composer.json
internalsystemerror: I thought the conflict section would take care of that? Say for instance, users on mezzio using a different container implementation, but still currently using laminas-form, this would require them to use laminas-servicemanager.

Adding the conflict for `<3.14.0` should prevent anyone from being able to use this latest version with an incompatible service manager, without specifically requiring it?

docs/book/v2/advanced.md
internalsystemerror: ! Thanks for pointing that out, I'll make sure to fix.

docs/book/v2/advanced.md
internalsystemerror: Will fix

internalsystemerror:
A couple of questions:
- Should this target a new major? I don't think so, but thought it worth asking.
- A lot of the psalm warnings were generated from the test folders and I managed to get rid of a few of them by calling static methods statically, but most are warnings that key might not exist in the array, which is exactly what's being tested so I added those to `psalm-baseline.xml`. Is that the best approach here? Or should we be improving the typing to make the tests redundant?

psalm-baseline.xml
```diff
@@ -436,6 +364,20 @@
...
+  <file src="test/Element/CheckboxTest.php">
+    <PossiblyInvalidMethodCall occurrences="1">
+      <code>getHaystack</code>
+    </PossiblyInvalidMethodCall>
+    <PossiblyUndefinedArrayOffset occurrences="1">
+      <code>$inputSpec['validators']</code>
+    </PossiblyUndefinedArrayOffset>
+    <UndefinedInterfaceMethod occurrences="1">
+      <code>getHaystack</code>
+    </UndefinedInterfaceMethod>
```
Ocramius, line 376: Inference regressed here: can we investigate why?

Ocramius (approved):
Besides worsened type inference, this is OK as-is, @internalsystemerror :+1:

src/Element/AbstractDateTime.php
```diff
@@ -20,6 +21,9 @@
 use function date;
 use function sprintf;
 
+/**
+ * @psalm-import-type ValidatorSpecification from InputFilterInterface
+ */
```
gsteel: This import doesn't appear to be used anywhere?

gsteel (changes requested):
Other than the unused import LGTM!

psalm-baseline.xml
internalsystemerror: This is just as I was saying above (afaik) that this is what is being tested for. So I'm not sure why these are showing now and not before (my theory: better types in the docblocks in src).

internalsystemerror:
Unused type import remove, rebased to the latest changes in `3.7.x`, psalm baseline updated and dependency lockfile updated too.

Should be ready for rereview now 👍

gsteel (approved):
LGTM!

Ocramius (approved):
Let's roll with it, thanks @internalsystemerror!

PR #180, opened 2022-07-25
Fixes Test Failures relating to InputFilter Regression and Doctrine Annotations Changes

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Doctrine Annotations now converts deprecations to exceptions (ref: https://github.com/doctrine/annotations/pull/438) but it's difficult to tell when the internal call to `trigger_error` for the deprecation fires vs the exception thrown from annotations, event though `convertDeprecationsToExceptions` is active here.

This pull also updates `inputfilter` to the latest release WRT https://github.com/laminas/laminas-inputfilter/pull/60

A failing test was occurring, due to other changes in InputFilter where factories are now fetching the `InputFilterManager` et al by FQCN rather than alias. This broke the `FormAbstractServiceFactoryTest` because of the contrived ServiceManager config. This shouldn't be an issue for consumers because they would be pulling services from a fully configured ServiceManger _(The aliases and services haven't changed)_

This should help other pulls to go green on latest deps

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
         "laminas/laminas-hydrator": "^4.3.1",
-        "laminas/laminas-inputfilter": "^2.13.0",
+        "laminas/laminas-inputfilter": "^2.19.1",
         "laminas/laminas-stdlib": "^3.7.1"
     },
     "conflict": {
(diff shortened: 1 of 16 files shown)


gsteel:
Probably going to need https://github.com/laminas/laminas-inputfilter/pull/60 before this will go green.

Note: relevant tests were passing in `~2.15.0` of `laminas-inputfilter`

Ocramius:
:ship: Thanks @gsteel!

PR #184, opened 2022-07-29
Fix regular expression matching date patterns in date/datetime `<select/>`-based helpers, allowing support for other locales

### LINKED ISSUE #160: Unescaped regex-meta-character in AbstractFormDateSelect
I think the regex-meta-character "-" (dash) in the date/time pattern should be escaped:
https://github.com/laminas/laminas-form/blob/beee25df568f1140d5553eabe51f97c6643f0464/src/View/Helper/AbstractFormDateSelect.php#L78

It should be:
```
"/([ \-,.\/]*(?:'[a-zA-Z]+')*[ \-,.\/]+)/"
```

Otherwise dates in the format `dd-MM-yy` are not recognized.
For example, the `IntlDateFormatter` returns for the `es_CL` locale that pattern.

mimmi20:
This will escape some meta characters.

Fixes #160

### DIFF
--- a/test/View/Helper/FormDateSelectTest.php
+++ b/test/View/Helper/FormDateSelectTest.php
@@ -9,6 +9,8 @@ use Laminas\Form\Element\DateSelect;
 use Laminas\Form\Element\Select;
 use Laminas\Form\Exception\DomainException;
 use Laminas\Form\View\Helper\FormDateSelect as FormDateSelectHelper;
+use ReflectionException;
+use ReflectionMethod;
 
 use function extension_loaded;
 
@@ -114,5 +116,29 @@ final class FormDateSelectTest extends AbstractCommonTestCase
         }
 
         $this->assertCount(31, $elements[0]->getValueOptions());
+    }
+
+    /**
+     * @group issue-160
+     * @throws ReflectionException
+     */
+    public function testParsePatternWithShortDateFormat(): void
+    {
+        $element  = new DateSelect('foo');
+        $dateType = IntlDateFormatter::SHORT;
+        $locale   = 'es_CL';
+
+        $this->helper->setLocale($locale);
+        $this->helper->setDateType($dateType);
+
+        $method = new ReflectionMethod($this->helper, 'parsePattern');
+        $method->setAccessible(true);
+
+        $pattern = $method->invoke($this->helper, $element->shouldRenderDelimiters());
+
+        $this->assertIsArray($pattern);
+        $this->assertArrayHasKey('day', $pattern);
+        $this->assertArrayHasKey('month', $pattern);
+        $this->assertArrayHasKey('year', $pattern);
     }
 }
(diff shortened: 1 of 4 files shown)


test/View/Helper/FormDateSelectTest.php
```diff
@@ -115,4 +115,32 @@ public function testGetElements(): void
...
+        $this->assertSame($dateType, $this->helper->getDateType());
+
+        $elements = $element->getElements();
+        $this->assertCount(3, $elements);
+
+        foreach ($elements as $subElement) {
+            $this->assertInstanceOf(Select::class, $subElement);
+        }
+
+        $this->assertCount(31, $elements[0]->getValueOptions());
```
Ocramius: I feel like this test isn't really readable: what are we testing here, specifically? 🤔

Ocramius (changes requested):
Few things to adjust:

 1. commit message does not explain what's going on - some more clarity needed there
 2. test case does not make the issue clear: if this test failed, I wouldn't understand why, nor what it represents

test/View/Helper/FormDateSelectTest.php
mimmi20: fixed

test/View/Helper/FormDateSelectTest.php
```diff
@@ -115,4 +117,28 @@ public function testGetElements(): void
...
+    public function testParsePatternWithShortDateFormat(): void
+    {
+        $element  = new DateSelect('foo');
+        $dateType = IntlDateFormatter::SHORT;
+        $locale   = 'es_CL';
+
+        $this->helper->setLocale($locale);
+        $this->helper->setDateType($dateType);
+
+        $method = new ReflectionMethod($this->helper, 'parsePattern');
```
Ocramius: Hmm, this looks even worse than before - we shouldn't test internal methods via reflection.

Also here, when reading this test out of context, it is unclear what you are verifying.

Here's my understanding:

 * when using `es_CL`
 * and setting a short date format
 * when I parse the pattern
 * then the pattern has day, month, year

So far, so good, but what's the effect on the public-facing behavior of this element helper?

Ocramius: As for the commit message: `laminas#160: escape regex-meta-character in AbstractFormDateSelect `still doesn't say much to the reader :|

mimmi20: I'm escaping two regex meta characters which are currently unescaped, like descriped in issue #160. I dont know how to test this without reflection as the functions which are causing the error are protected.

Xerkus: This should be tested via public behavior. If there is no public behavior that exposes the bug then it is fine to fix without providing a test.

Ocramius: > then it is fine to fix without providing a test.

No, it is not: it means we have dead/unreachable code then, and it should just be dropped or refactored instead.
Bugfixing without a valid test is just more maintenance burden for upstream maintainers.

Ocramius: Checking this out (locally) to re-design the test, and try to trip the issue with `es_CL`.

Ocramius: I've pushed faefb5bbe578a071b95f3a34a145940a28b2acee, which solves this discussion, and provides some context of what is going on (for posterity).

Meanwhile, thanks for the patience, @mimmi20!

PR #186, opened 2022-08-04
adding support of passwordrules option to password element rendering

necromant2005:
Adding support of passwordrules option for password element based on apple and 1password recommendation https://developer.apple.com/password-rules/ 

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

1password and Apple support options for automatic password generation for this they are using option password rules
e.q. 
`<input type="password" passwordrules="minlength: 8; maxlength: 36; required: lower; required: upper; required: digit; required: [-];">` 
it will allow automatically generate password like: "suhxe6-buwBum-zakvab", "hegxop-6ranqa-dIzkem" 
All of this provides easier path for users who struggle to generate suitable password 

- added option "passwordrules" to attributes supported by View/Helper/Password 
- in order to reproduce build element password with password rules attribute
- expectation: view/helper/password renders option passwordrules appear
- tests to cover this case - added

### DIFF
--- a/src/View/Helper/FormPassword.php
+++ b/src/View/Helper/FormPassword.php
@@ -14,20 +14,21 @@ class FormPassword extends FormInput
      * @var array
      */
     protected $validTagAttributes = [
-        'name'         => true,
-        'autocomplete' => true,
-        'autofocus'    => true,
-        'disabled'     => true,
-        'form'         => true,
-        'maxlength'    => true,
-        'minlength'    => true,
-        'pattern'      => true,
-        'placeholder'  => true,
-        'readonly'     => true,
-        'required'     => true,
-        'size'         => true,
-        'type'         => true,
-        'value'        => true,
+        'name'          => true,
+        'autocomplete'  => true,
+        'autofocus'     => true,
+        'disabled'      => true,
+        'form'          => true,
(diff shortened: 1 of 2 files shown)


Slamdunk:
To be honest, I'm not enthusiast about this: https://github.com/mozilla/standards-positions/pull/402/files

froschdesign:
Even though I really like the fact that the possible attributes are kept up to date, but the current status of the `passwordrules` attribute is only a **proposal**. (See: https://github.com/whatwg/html/issues/3518)

If the attribute is needed then [it can be added as a valid attribute to the view helper](https://docs.laminas.dev/laminas-form/v3/helper/abstract-helper/#allowing-framework-specific-attributes). A factory or delegator can help here.

necromant2005:
Safari is already support it, other browsers are coming

Slamdunk:
> other browsers are coming

Mozilla considers it harmful as per link https://github.com/mozilla/standards-positions/pull/402/files :\

Ocramius:
Can totally stay behind Mozilla's choice here.

PR #189, opened 2022-10-13
Add support for PHP 8.2, remove support for PHP 7.4, remove prophecy usage

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     "homepage": "https://laminas.dev",
     "license": "BSD-3-Clause",
     "require": {
-        "php": "^7.4 || ~8.0.0 || ~8.1.0",
+        "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
         "laminas/laminas-hydrator": "^4.3.1",
         "laminas/laminas-inputfilter": "^2.19.1",
         "laminas/laminas-stdlib": "^3.7.1"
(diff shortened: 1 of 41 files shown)


Ocramius (approved):
Excellent, thanks @gsteel!

Ocramius:
Ah, I see @Slamdunk is handling this 👍

composer.json
```diff
@@ -25,24 +25,23 @@
     "require-dev": {
         "ext-intl": "*",
         "doctrine/annotations": "^1.13.3",
-        "laminas/laminas-captcha": "^2.11.0",
```
Slamdunk, line 28: Can you update the `conflict` section as well please?

Slamdunk (changes requested):
Just a minor adjustment to `composer.json` and we can go :muscle:

Slamdunk (approved):

composer.json
gsteel: Assuming you want the conflicts to prevent installation of anything less than most recent? Why is this necessary?

Slamdunk: Only tested versions are supported.
Without the `conflict` section:

1. a user may install an unsupported version
2. find a bug
3. report an Issue here
4. we close it with "Bump your deps"

`conflict` prevents this all

gsteel: Thanks for explaining @Slamdunk 👍

PR #190, opened 2022-10-14
Improve type inference for FormElementManager

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### DIFF
--- a/src/FormElementManager.php
+++ b/src/FormElementManager.php
@@ -22,6 +22,8 @@ use function sprintf;
  * laminas-servicemanager v3-compatible plugin manager implementation for form elements.
  *
  * Enforces that elements retrieved are instances of ElementInterface.
+ *
+ * @extends AbstractPluginManager<ElementInterface>
  */
 class FormElementManager extends AbstractPluginManager
 {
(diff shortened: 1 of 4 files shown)


Ocramius (approved):
LGTM 👍

src/FormElementManager.php
```diff
@@ -22,6 +22,8 @@
  * laminas-servicemanager v3-compatible plugin manager implementation for form elements.
  *
  * Enforces that elements retrieved are instances of ElementInterface.
+ *
+ * @extends AbstractPluginManager<ElementInterface>
```
Slamdunk, line 26: May I ask you why here the `@extends` is simple, while on `laminas-validator` [you opted](https://github.com/laminas/laminas-validator/commit/e85cc0fb2e2e28bd1e75124d3d262a6ee2b995e7#diff-70f7b0f6ed7c22fbb40b0ff2ddc6e0115328d733bdc6f6435893ee66c62ceb04R57) for
```
 * @template InstanceType of ValidatorInterface
 * @extends AbstractPluginManager<InstanceType>
```
?

gsteel: I think validator should have taken the same approach, i.e. not define a custom type. Custom types are useful, for example, in View and Filter where the type can be `HelperInterface|callable()|SomethingElse` - in this case it's only ever going to be an ElementInterface so the custom type was unnecessary.

Slamdunk (approved):

Slamdunk:
Thank you @gsteel

PR #193, opened 2022-10-14
Retrieve dependencies by FQCN where appropriate

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

This patch updates all `$container->get()` call sites to use a FQCN instead of a string alias. DI configuration is updated so that factories are registered against FQCN but maintains existing service aliases.

### DIFF
--- a/src/ConfigProvider.php
+++ b/src/ConfigProvider.php
@@ -34,14 +34,14 @@ final class ConfigProvider
                 FormAbstractServiceFactory::class,
             ],
             'aliases'            => [
-                Annotation\AnnotationBuilder::class => 'FormAnnotationBuilder',
-                Annotation\AttributeBuilder::class  => 'FormAttributeBuilder',
-                FormElementManager::class           => 'FormElementManager',
+                'FormAnnotationBuilder' => Annotation\AnnotationBuilder::class,
+                'FormAttributeBuilder'  => Annotation\AttributeBuilder::class,
+                'FormElementManager'    => FormElementManager::class,
             ],
             'factories'          => [
-                'FormAnnotationBuilder' => Annotation\BuilderAbstractFactory::class,
-                'FormAttributeBuilder'  => Annotation\BuilderAbstractFactory::class,
-                'FormElementManager'    => FormElementManagerFactory::class,
+                Annotation\AnnotationBuilder::class => Annotation\BuilderAbstractFactory::class,
+                Annotation\AttributeBuilder::class  => Annotation\BuilderAbstractFactory::class,
+                FormElementManager::class           => FormElementManagerFactory::class,
             ],
         ];
     }
--- a/src/ConfigProvider.php
+++ b/src/ConfigProvider.php
@@ -34,14 +34,14 @@ final class ConfigProvider
                 FormAbstractServiceFactory::class,
             ],
             'aliases'            => [
-                'FormAnnotationBuilder' => Annotation\AnnotationBuilder::class,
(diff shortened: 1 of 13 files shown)


src/ConfigProvider.php
```diff
@@ -34,14 +34,14 @@ public function getDependencyConfig(): array
...
+                'FormAttributeBuilder'  => Annotation\AttributeBuilder::class,
+                'FormElementManager'    => FormElementManager::class,
             ],
             'factories'          => [
-                'FormAnnotationBuilder' => Annotation\BuilderAbstractFactory::class,
-                'FormAttributeBuilder'  => Annotation\BuilderAbstractFactory::class,
-                'FormElementManager'    => FormElementManagerFactory::class,
+                Annotation\AnnotationBuilder::class => Annotation\BuilderAbstractFactory::class,
+                Annotation\AttributeBuilder::class  => Annotation\BuilderAbstractFactory::class,
+                FormElementManager::class           => FormElementManagerFactory::class,
```
Ocramius: Note here: swapping aliases and factories has a secondary effect of messing with delegators.

I am not 100% sure if delegators operate the same way on factories or aliases. Consider keeping the priority as it was before :)

Ocramius (changes requested):

src/ConfigProvider.php
Slamdunk: I asked him to do so: I'd rather check the delegator issue than keeping this forever

Slamdunk: Reported in https://github.com/laminas/laminas-servicemanager/issues/158

I hope to have a PR there in the next week

Ocramius: This is still a BC break then: IMO not worth it for now, or to be extracted and targeted (just this bit) for `4.0.0`

Slamdunk: It's a BC break only if the current `laminas-servicemanager` major doesn't work with both alias and aliased name

Ocramius: If you manage to fix it, we can bump `laminas-servicemanager` to the fix version from laminas/laminas-servicemanager#158 and do it in a minor release.

Removing milestone here meanwhile, since it will be blocked for another while.

gsteel: Why don't I swap the aliases back for now - the rest can be merged preventing future conflicts?

IIRC, delegators simply don't work on aliases and never will: https://github.com/laminas/laminas-servicemanager/pull/91

Ocramius: Yeah, I think that would be OK to do meanwhile :+1:

Ocramius: @Slamdunk your call here - can also discuss it tonight.

src/Module.php
```diff
@@ -28,12 +29,13 @@ public function getConfig(): array
      */
     public function init(ModuleManager $moduleManager): void
     {
-        $event           = $moduleManager->getEvent();
-        $container       = $event->getParam('ServiceManager');
+        $event     = $moduleManager->getEvent();
+        $container = $event->getParam('ServiceManager');
```
Ocramius, line 33: Is this using single quotes because the service is defined outside here?

Ocramius (approved):
LGTM 👍

src/Module.php
gsteel: I think this is defined by module manager, somewhere in MVC

Ocramius:
Thanks @gsteel!

PR #195, opened 2022-10-26
Updates description for laminas-mvc application integration to simplify the description and usage

froschdesign:
Creating a separate factory for the controller is too complicated, because laminas-servicemanager offers an easy way to reach the goal.

### DIFF
--- a/docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
+++ b/docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
@@ -25,7 +25,7 @@ use Laminas\Form\Form;
 use Laminas\InputFilter\InputFilterProviderInterface;
 use Laminas\Validator\StringLength;
 
-class AlbumForm extends Form implements InputFilterProviderInterface
+final class AlbumForm extends Form implements InputFilterProviderInterface
 {
     public function init() : void
     {
@@ -78,7 +78,9 @@ use Album\Form\AlbumForm;
 use Laminas\Form\FormElementManager;
 use Laminas\Mvc\Controller\AbstractActionController;
 
-class AlbumController extends AbstractActionController
+use function assert;
+
+final class AlbumController extends AbstractActionController
 {
     public function __construct(
         public readonly FormElementManager $formElementManager
@@ -87,6 +89,7 @@ class AlbumController extends AbstractActionController
     public function addAction()
     {
         $form = $this->formElementManager->get(AlbumForm::class);
+        assert($form instanceof AlbumForm);
     
         // Set action attribute
         $form->setAttribute(
--- a/docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
+++ b/docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
@@ -67,80 +67,48 @@ class AlbumForm extends Form implements InputFilterProviderInterface
 }
 ```
 
-## Using Form
+## Create Controller
 
-### Create Controller
-
-[Create a controller class](https://docs.laminas.dev/laminas-mvc/quick-start/#create-a-controller)
-and inject the form via the constructor, e.g.
-`module/Album/Controller/AlbumController.php`:
+[Create a controller class](https://docs.laminas.dev/laminas-mvc/quick-start/#create-a-controller) and inject the form element manager via the constructor, e.g. `module/Album/Controller/AlbumController.php`:
 
 ```php
 namespace Album\Controller;
 
 use Album\Form\AlbumForm;
-use Laminas\Form\FormInterface;
+use Laminas\Form\FormElementManager;
 use Laminas\Mvc\Controller\AbstractActionController;
 
 class AlbumController extends AbstractActionController
 {
-    /** @var FormInterface */
-    private $form;
-    
-    public function __construct(AlbumForm $form)
-    {
-        $this->form = $form;
-    }
+    public function __construct(
+        public readonly FormElementManager $formElementManager
+    ) {}
     
     public function addAction()
     {
+        $form = $this->formElementManager->get(AlbumForm::class);
+    
         // Set action attribute
-        $this->form->setAttribute(
+        $form->setAttribute(
             'action',
             $this->url()->fromRoute('album', ['action' => 'add'])
         );
 
-        $variables = ['form' => $this->form];
+        $variables = ['form' => $form];
         
         if (! $this->getRequest()->isPost()) {
             return $variables;
         }
 
         // Validation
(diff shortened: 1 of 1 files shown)


docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -67,48 +67,42 @@ class AlbumForm extends Form implements InputFilterProviderInterface
...
-    /** @var FormInterface */
-    private $form;
-    
-    public function __construct(AlbumForm $form)
-    {
-        $this->form = $form;
-    }
+    public function __construct(
+        public readonly FormElementManager $formElementManager
+    ) {}
```
weierophinney, line 87: I wonder if instead we could do the following

```php
    private AlbumForm $form;

    public function __construct(FormElementManager $formElementManager)
    {
        $this->form = $formElementManager->get(AlbumForm::class);
    }
```

This would ensure that we have a form of the expected type, and it would provide a slight performance benefit if the controller were re-used (e.g., if using something like OpenSwoole), as we wouldn't be pulling the form from the form element manager each time.

weierophinney (commented):
Overall, this is fine; I just wonder if maybe the controller constructor should pull the form instance and memoize it, instead of storing the form element manager.

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: Good idea!

kynx: @weierophinney maybe I misunderstand, but... forms are just dripping with state. OK, this is documenting MVC-land, but I would still encourage people to get a fresh one every time they need it, if just to encourage right thinking.

I like the idea of injecting `FormElementManager` though. Maybe a `private function getForm(): AlbumForm` could ensure the type?

Slamdunk: I agree with @kynx : a Form is neither a `Service` nor a `Value Object`. A Form is an `Entity`, stateful and mutable.
In a well structured application this shouldn't matter much as a Controller should be unique and single-purpose, but still the idea of having it as a constructor dependency feels wrong to me.
In my point of view, the perfect Controller have only `Service` dependencies, injected in the constructor and saved into `readonly` attributes; `Entities` have to be created where and when needed.

froschdesign: @Slamdunk 
> `Entities` have to be created where and when needed.

What is your concrete proposal? `$form = new AlbumForm();`?

Slamdunk:
```php
final class MyController implements RequestHandlerInterface
{
    public function __construct(
        private readonly FormElementManager $formElementManager,
    ) {
    }

    public function handle(ServerRequestInterface $request): ResponseInterface
    {
        // Some code
        
        $form = $this->formElementManager->get(AlbumForm::class);
        
        // Other code
    }
}
```

froschdesign: @Slamdunk  
Very good, I thought you were going to skip the from element manager. 😄

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
```diff
@@ -67,48 +67,43 @@ class AlbumForm extends Form implements InputFilterProviderInterface
...
     
     public function addAction()
     {
+        /** @var AlbumForm $form */
+        $form = $this->formElementManager->get(AlbumForm::class);
```
Slamdunk: See https://github.com/doctrine/coding-standard/pull/47:
```suggestion
        $form = $this->formElementManager->get(AlbumForm::class);
        \assert($form instanceof AlbumForm);
```

Slamdunk (changes requested):

docs/book/v3/application-integration/usage-in-a-laminas-mvc-application.md
froschdesign: _We do not use the Doctrine coding standard._

That is correct, but the question is if it belongs in this tutorial? The one line raises more questions and therefore needs additional explanation. The problem is that we use it in the code of various components, but nowhere in the documentation.

Slamdunk: > _We do not use the Doctrine coding standard._

Neither are `readonly` attributes. My link was just about the *why* we should do it.

> The one line raises more questions and therefore needs additional explanation.

We have never explained `/** @var */` either.

> The problem is that we use it in the code of various components, but nowhere in the documentation.

`assert()` in the worst case scenario behaves exactly like `/** @var */`; in the best case scenario is a runtime type check.
Why not pushing new good practices over the worse old alternative?

froschdesign: > We have never explained /** @var */ either.

This is nothing new in the world of PHP and also in the documentation.

> Why not pushing new good practices over the worse old alternative?

Please don't misunderstand me, I would like to take it over but we are still fighting and have to live with "worse old alternative" as well. Because we have to deal with the old, ugly and wrong description in the getting started tutorial. Therefore, a balance must be struck here.

_Give me some time to consider._

froschdesign: @Slamdunk 
Thanks for your review and input! 👍🏻

froschdesign: @Slamdunk 
I knew I read it somewhere. 😃 

We already have this topic in the contributing guide:

> ### Performing assertions on the code
> We distinguish assertions in two categories:
>
> * Expectations of something that is always supposed to be true
> * Expectations of some input to follow some rule
>
> The former is a development-only concern, aiming to aid tools (e.g. IDEs, static analysis) to perform a more accurate type detection. In this scenario, you should use the [assert()](https://www.php.net/manual/en/function.assert.php) function, which executes the expression when [zend.assertions](https://www.php.net/manual/en/ini.core.php#ini.zend.assertions) is enabled. That will make sure that the condition is valid, while easing future refactoring.
>
> …

https://github.com/laminas/.github/blob/main/CONTRIBUTING.md#performing-assertions-on-the-code

Slamdunk: Nice, so we can use it :muscle:

Slamdunk (approved):

Slamdunk:
Thank you @froschdesign

PR #207, opened 2023-03-11
Fixes #22: "ComposedObject" Annotation does not add inputSpec to element configuration

### LINKED ISSUE #22: "ComposedObject" Annotation does not add inputSpec to element configuration
I noticed, that my form was not validating correctly. After some digging in code, I found, that although inputSpec is properly set in ElementAnnotationsListener::handleComposedObjectAnnotation, it gets ignored in AnnotationBuilder::configureElement

So here is what I found was causing it, and how I fixed it (although I am not sure, if that is correct):
AnnotationBulder.php:

``` php
    protected function configureElement($annotations, $reflection, $formSpec, $filterSpec)
    {
        // If the element is marked as exclude, return early
        if ($this->checkForExclude($annotations)) {
            return;
        }

        $events = $this->getEventManager();
        $name   = $this->discoverName($annotations, $reflection);

        $elementSpec = new ArrayObject([
            'flags' => [],
            'spec'  => [
                'name' => $name
            ],
        ]);
        $inputSpec = new ArrayObject([
            'name' => $name,
        ]);

        $params = [
            'name'        => $name,
            'elementSpec' => $elementSpec,
            'inputSpec'   => $inputSpec,  // Before the fix, this would not change its vallue
            'formSpec'    => $formSpec,
            'filterSpec'  => $filterSpec,
        ];
        foreach ($annotations as $annotation) {
            $params['annotation'] = $annotation;
// -> This causes ElementAnnotationsListener::handleComposedObjectAnnotation to configure the element
            $events->trigger(__FUNCTION__, $this, $params);
        }
        // Rest of the code is omitted (not relevant)
    }
```

ElementAnnotationsListener.php:

``` php
public function handleComposedObjectAnnotation($e)
    {
        $annotation = $e->getParam('annotation');
        if (!$annotation instanceof ComposedObject) {
            return;
        }

        $class             = $annotation->getComposedObject();
        $annotationManager = $e->getTarget();
        $specification     = $annotationManager->getFormSpecification($class);

        $name        = $e->getParam('name');
        $elementSpec = $e->getParam('elementSpec');

        if ($annotation->isCollection()) {
// Omitted, not relevant
        } else {
            // Compose input filter into parent input filter
            $inputFilter = $specification['input_filter'];
            if (!isset($inputFilter['type'])) {
                $inputFilter['type'] = 'Zend\InputFilter\InputFilter';
            }

// So here is what causes the problem:
// For some reason setting inputSpec like this causes it to be ignored in AnnotationBuilder::configureElement
//            $e->setParam('inputSpec', $inputFilter);
// This seems to work, now the composed object receives the correct inputSpec
            /** @var ArrayObject $inputSpec */
            $inputSpec = $e->getParam('inputSpec');
            $inputSpec->exchangeArray($inputFilter);

            unset($specification['input_filter']);
// Rest of the code is omitted, not relevant
        }
    }
```

I don't believe that this is expected behavior, because every other aspect of the element configuration can be set in ElementAnnotationsListener::handleComposedObjectAnnotation. 

---

Originally posted by @prendit at https://github.com/zendframework/zend-form/issues/125

riul88:
Signed-off-by: Raul Robledo <riul88@gmail.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes/no
| New Feature   | no
| RFC           | yes/no
| QA            | yes/no

### Description
Fix suggested by @weierophinney 
Issue reproducible by unit test testAllowsComposingChildEntities but the test case didn't validate the filters or validators from the child entity, the unit test was updated as part of this PR to perform the missing assertions to ensure the element validators and filters are configured

The changes to unit test testAllowsComposingMultipleChildEntities which is used when ComposedObject is marked as isCollection are only preventive not impacted by the code change in the PR

### DIFF
--- a/test/Annotation/AbstractBuilderTestCase.php
+++ b/test/Annotation/AbstractBuilderTestCase.php
@@ -6,6 +6,7 @@ namespace LaminasTest\Form\Annotation;
 
 use ArrayObject;
 use Generator;
+use Laminas\Filter\StringTrim;
 use Laminas\Form\Annotation;
 use Laminas\Form\Element;
 use Laminas\Form\Element\Collection;
@@ -17,6 +18,9 @@ use Laminas\InputFilter\Input;
 use Laminas\InputFilter\InputFilterInterface;
 use Laminas\InputFilter\InputInterface;
 use Laminas\Stdlib\PriorityList;
+use Laminas\Validator\EmailAddress;
+use Laminas\Validator\NotEmpty;
+use Laminas\Validator\StringLength;
 use LaminasTest\Form\TestAsset;
 use LaminasTest\Form\TestAsset\Annotation\Entity;
 use LaminasTest\Form\TestAsset\Annotation\Form;
@@ -215,6 +219,23 @@ abstract class AbstractBuilderTestCase extends TestCase
         self::assertInstanceOf(InputFilterInterface::class, $composed);
         self::assertTrue($composed->has('username'));
         self::assertTrue($composed->has('password'));
+        $usernameInput = $composed->get('username');
+        $usernameValidators = $usernameInput->getValidatorChain()->getValidators();
+        self::assertCount(2, $usernameValidators);
+        self::assertInstanceOf(NotEmpty::class, $usernameValidators[0]['instance']);
+        self::assertInstanceOf(StringLength::class, $usernameValidators[1]['instance']);
+        $usernameFilters = $usernameInput->getFilterChain()->getFilters()->toArray();
+        self::assertCount(1, $usernameFilters);
+        self::assertInstanceOf(StringTrim::class, $usernameFilters[0]);
+
+        $passwordInput = $composed->get('password');
+        $passwordValidators = $passwordInput->getValidatorChain()->getValidators();
+        self::assertCount(1, $passwordValidators);
+        self::assertInstanceOf(EmailAddress::class, $passwordValidators[0]['instance']);
+
+        $passwordFilters = $passwordInput->getFilterChain()->getFilters()->toArray();
+        self::assertCount(1, $passwordFilters);
+        self::assertInstanceOf(StringTrim::class, $passwordFilters[0]);
     }
 
     public function testAllowsComposingMultipleChildEntities(): void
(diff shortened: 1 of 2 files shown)


test/Annotation/AbstractBuilderTestCase.php
```diff
@@ -215,6 +221,28 @@ public function testAllowsComposingChildEntities(): void
...
+        $passwordInput = $composed->get('password');
+        self::assertInstanceOf(Input::class, $passwordInput);
+        $validatorChain     = $passwordInput->getValidatorChain();
+        $passwordValidators = $validatorChain->getValidators();
+        self::assertCount(1, $passwordValidators);
+        self::assertInstanceOf(EmailAddress::class, $passwordValidators[0]['instance']);
+
+        $passwordFilters = $passwordInput->getFilterChain()->getFilters()->toArray();
+        self::assertCount(1, $passwordFilters);
+        self::assertInstanceOf(StringTrim::class, $passwordFilters[0]);
```
Ocramius, line 245: Can we split it out to a smaller test highlighting just what is being fixed? :thinking:

Ocramius (changes requested):

test/Annotation/AbstractBuilderTestCase.php
Slamdunk: Without the fix, `username` and `password` result indeed with no validator at all, wrongfully.

test/Annotation/AbstractBuilderTestCase.php
```diff
@@ -231,6 +259,47 @@ public function testAllowsComposingMultipleChildEntities(): void
...
+                '0' => [
+                    'name' => 'StringTrim',
+                ],
+            ],
+            'validators' => [
+                '0' => [
+                    'name' => 'EmailAddress',
+                ],
+            ],
+        ], $passwordFilterSpec);
```
Slamdunk, line 302: This test passes even without the fix though, but I like it has been added.

Slamdunk (approved):

weierophinney (approved):
:ship:

@Slamdunk Since you're official maintainer for this component now, I'll let you merge.

Slamdunk:
Thank you @riul88

PR #212, opened 2023-03-22
Fixes #33 Collection has empty bonded nested objects

### LINKED ISSUE #33: Collection has empty bonded nested objects
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

---

Original Issue: https://api.github.com/repos/zendframework/zendframework/issues/7156
User: @epoleacov
Created On: 2015-01-25T22:22:49Z
Updated At: 2015-10-19T21:10:08Z
Body
When nested object bonded to form with related collection elements, this collection elements has just empty objects.
So for example if we run next test:

```
ZendTest\Form\Element\CollectionTest::testCanBindObjectMultipleNestedFieldsets()
```

final object used for assertion should have original data ($_product->getObject()->getPrice() == 200), but it has just NULL instead.

Main cause of that is because when values extracted from collection is just takes extracted values, but forget about objects what was used in extraction which should be saved inside of collection under "object" property, which on his side used later for filling collection target elements on populateValues();

Next is line where values extracted, but objects are lost:
https://github.com/zendframework/zf2/blob/b1e9f3bb3ce9cdc925139d98672af40f6b0f33d7/library/Zend/Form/Element/Collection.php#L528

---
### Comment

User: @iprdp
Created On: 2015-04-20T18:15:00Z
Updated At: 2015-04-20T18:15:00Z
Body
I have worked around this bug by creating the targetElementInstance during extract(). Not sure if this is good enough for a fix.

---
### Comment

User: @adamlundrigan
Created On: 2015-10-19T21:10:08Z
Updated At: 2015-10-19T21:10:08Z
Body
@epoleacov could you provide an example showing the behaviour you are seeing?  I'm not having any luck reproducing what you're trying to explain

---

---

Originally posted by @GeeH at https://github.com/zendframework/zend-form/issues/95

riul88:
Update to include target type in collection element spec built from annotation to later be used to create new element

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes/no
| New Feature   | no
| RFC           | yes/no
| QA            | yes

### Description

As described in issue #33 , collection has empty bounded nested objects, while iterating thru the data the method  createNewTargetElementInstance is called to retrieve a new instance of the element, but as the Object is null when trying to hydrate the element the result is an array of nulls in the collection

Added test case testAllowsComposingMultipleChildEntitiesWithEntityBind to validate the result object after binding, with current functionality the assertions to ensure the elements are instances of EntityObjectPropertyHydrator fail as the values of the elements is null

### DIFF
--- a/src/Element/Collection.php
+++ b/src/Element/Collection.php
@@ -9,6 +9,7 @@ use Laminas\Form\Exception;
 use Laminas\Form\Fieldset;
 use Laminas\Form\FieldsetInterface;
 use Laminas\Form\FormInterface;
+use Laminas\Form\InputFilterProviderFieldset;
 use Laminas\Stdlib\ArrayUtils;
 use Laminas\Stdlib\Exception\InvalidArgumentException;
 use Traversable;
@@ -533,9 +534,11 @@ class Collection extends Fieldset
     protected function createNewTargetElementInstance(): ElementInterface
     {
         assert($this->targetElement !== null);
-        if ($this->targetElement->getOption('target_type')) {
-            $type = $this->targetElement->getOption('target_type');
-            $this->targetElement->setObject(new $type());
+        if ($this->targetElement instanceof InputFilterProviderFieldset) {
+            if ($this->targetElement->getOption('target_type')) {
+                $type = $this->targetElement->getOption('target_type');
+                $this->targetElement->setObject(new $type());
+            }
         }
         return clone $this->targetElement;
     }
--- a/src/Element/Collection.php
+++ b/src/Element/Collection.php
@@ -15,11 +15,13 @@ use Laminas\Stdlib\Exception\InvalidArgumentException;
 use Traversable;
 
 use function assert;
+use function class_exists;
 use function count;
 use function gettype;
 use function is_array;
 use function is_int;
 use function is_object;
+use function is_string;
 use function iterator_to_array;
 use function max;
 use function sprintf;
@@ -535,8 +537,8 @@ class Collection extends Fieldset
     {
         assert($this->targetElement !== null);
         if ($this->targetElement instanceof InputFilterProviderFieldset) {
-            if ($this->targetElement->getOption('target_type')) {
-                $type = $this->targetElement->getOption('target_type');
+            if (null !== ($type = $this->targetElement->getOption('target_type'))) {
+                assert(is_string($type) && class_exists($type));
                 $this->targetElement->setObject(new $type());
             }
         }
--- a/src/Element/Collection.php
+++ b/src/Element/Collection.php
@@ -163,7 +163,7 @@ class Collection extends Fieldset
      * Set the object used by the hydrator
      * In this case the "object" is a collection of objects
      *
-     * @param  iterable $object
+     * @param iterable $object
      * @return $this
      * @throws Exception\InvalidArgumentException
      */
@@ -533,6 +533,10 @@ class Collection extends Fieldset
     protected function createNewTargetElementInstance(): ElementInterface
     {
         assert($this->targetElement !== null);
+        if ($this->targetElement->getOption('target_type')) {
+            $type = $this->targetElement->getOption('target_type');
+            $this->targetElement->setObject(new $type());
+        }
         return clone $this->targetElement;
     }
 
--- a/test/Annotation/AbstractBuilderTestCase.php
+++ b/test/Annotation/AbstractBuilderTestCase.php
@@ -299,7 +299,7 @@ abstract class AbstractBuilderTestCase extends TestCase
         self::assertInstanceOf(EntityObjectPropertyHydrator::class, $entity->child[1]);
     }
 
-    protected function validateEntityFilterSpec(array $filterSpec)
+    protected function validateEntityFilterSpec(array $filterSpec): void
     {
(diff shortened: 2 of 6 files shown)


src/Element/Collection.php
```diff
@@ -533,6 +534,12 @@ public function extract(): array
     protected function createNewTargetElementInstance(): ElementInterface
     {
         assert($this->targetElement !== null);
+        if ($this->targetElement instanceof InputFilterProviderFieldset) {
+            if ($this->targetElement->getOption('target_type')) {
+                $type = $this->targetElement->getOption('target_type');
```
Slamdunk:
```suggestion
            if (null !== ($type = $this->targetElement->getOption('target_type'))) {
                assert(is_string($type) && class_exists($type));
```

test/Annotation/AbstractBuilderTestCase.php
```diff
@@ -261,7 +263,44 @@ public function testAllowsComposingMultipleChildEntities(): void
...
+        self::assertNull($entity->child);
+        $form->bind($entity);
+        $form->setData([
+            'child' => [
+                '0' => ['password' => 'email@test.com', 'username' => 'user'],
+                '1' => ['password' => 'email2@test.com', 'username' => 'user2'],
+            ],
+        ]);
+        self::assertTrue($form->isValid());
+        self::assertNotNull($entity->child);
```
Slamdunk, line 294: This PSalm `DocblockTypeContradiction` can be added to the baseline:
```console
vendor/bin/psalm --no-cache --set-baseline=psalm-baseline.xml --update-baseline
```

test/TestAsset/Annotation/EntityComposingMultipleEntitiesObjectPropertyHydrator.php
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @Annotation\Name("hierarchical")
+ * @Annotation\Hydrator("Laminas\Hydrator\ObjectPropertyHydrator")
+ */
+#[Annotation\Name("hierarchical")]
+#[Annotation\Hydrator(ObjectPropertyHydrator::class)]
+class EntityComposingMultipleEntitiesObjectPropertyHydrator
+{
+    /**
+     * @var null|Entity
```
Slamdunk:
```suggestion
     * @var null|list<EntityObjectPropertyHydrator>
```

test/TestAsset/Annotation/EntityComposingMultipleEntitiesObjectPropertyHydrator.php
```diff
@@ -0,0 +1,28 @@
...
+     * @Annotation\ComposedObject("LaminasTest\Form\TestAsset\Annotation\EntityObjectPropertyHydrator", isCollection=true)
+     */
+    #[Annotation\ComposedObject(
+        EntityObjectPropertyHydrator::class,
+        isCollection: true,
+        options: ['create_new_objects' => true]
+    )]
+    public $child;
```
Slamdunk:
```suggestion
    public ?array $child = null;
```

Slamdunk (changes requested):
Overall looks good, just few adjustments and we can ship it :muscle:

froschdesign:
@Slamdunk 
For a bugfix the target branch is wrong.

Slamdunk:
Thank you @froschdesign , I picked the right milestone but I miss the target branch.

I wish we could automatically detect branch/milestone mismatches :thinking:

froschdesign:
@Slamdunk 
Unfortunately, the explanation of which branch should be used for bug fixes or new features is too far away. So nobody read or can read this: https://github.com/laminas/.github/blob/a1debce9a0842e3000ae9e01fbc60e32db62901d/CONTRIBUTING.md#what-branch-to-issue-the-pull-request-against
(The link to the contribution guidelines, for example, is located below each text field.)
And GitHub [uses the default branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

> I wish we could automatically detect branch/milestone mismatches 

But I think that leads to other problems.

src/Element/Collection.php
riul88: Implemented, thanks!

test/Annotation/AbstractBuilderTestCase.php
riul88: File updated, thanks!

test/TestAsset/Annotation/EntityComposingMultipleEntitiesObjectPropertyHydrator.php
riul88: Done, thanks!

test/TestAsset/Annotation/EntityComposingMultipleEntitiesObjectPropertyHydrator.php
riul88: Done, thanks!

riul88:
@Slamdunk thanks so much for all the feedback, I added a commit with the refactoring and fixes

Slamdunk (approved):

PR #218, opened 2023-06-02
Include exclamation point in month element format to prevent time-of-day related bugs

### LINKED ISSUE #217: Bug of day 31 (Element\Month)
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | any

#### Summary

When using a month element with the server date May 31st for example, the default validation of the DateStep returns false when selecting a month with less than 31 days.

#### Current behavior

"The input is not a valid step"

#### How to reproduce

Adjust the date to 2023-05-31

```php
use Laminas\Form\Form;
use Laminas\Form\Element;

$form = new Form();
$mesAnoIni = new Element\Month('mes_ano_inicio');
$form->add($mesAnoIni);
$form->setData(['mes_ano_inicio' => '2023-02']);
// $mesAnoIni->setAttribute('step', 'any'); // temporary fix
var_dump($form->isValid()); // return false
```

#### Expected behavior

Date validated regardless of the day on the server

#### How to permanently fix

[https://www.php.net/manual/en/datetimeimmutable.createfromformat.php] datetimeimmutable.createfromformat.php

All fields are initialized with the current date/time. In that case, you'll want to reset them to "zero". You do this by including the ! character as the first character in your format, or | as your last.

[src/Element/Month.php](https://github.com/laminas/laminas-form/blob/3.11.x/src/Element/Month.php)
```php
...
class Month extends AbstractDateTime
{
    /**
     * A valid format string accepted by date()
     *
     * @var string
     */
    protected $format = '!Y-m';`
```

samuelgmartins:
Update to include exclamation mark in month element format

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description
As described in issue #217, the month element when validated with DateStep on the 31st and with a month value of less than 31 days returns false validation.

Added testTestDayTimeReset test case to validate if month element date is reset

Fixes #217

### DIFF
--- a/test/Element/MonthTest.php
+++ b/test/Element/MonthTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace LaminasTest\Form\Element;
 
 use DateInterval;
+use DateTime;
 use Laminas\Form\Element\Month as MonthElement;
 use Laminas\Validator\DateStep;
 use Laminas\Validator\GreaterThan;
@@ -80,5 +81,12 @@ final class MonthTest extends TestCase
         self::assertArrayHasKey('validators', $inputSpec);
         $monthValidator = $inputSpec['validators'][0];
         self::assertEquals($expected, $monthValidator->isValid($value));
+    }
+
+    public function testTestDayTimeReset(): void
+    {
+        $element = new MonthElement('foo');
+        $date    = DateTime::createFromFormat($element->getFormat(), '2023-01');
+        self::assertEquals($date->format('d H:i:s'), '01 00:00:00');
     }
 }
--- a/src/Element/Month.php
+++ b/src/Element/Month.php
@@ -16,7 +16,7 @@ class Month extends AbstractDateTime
      *
      * @var string
      */
-    protected $format = 'Y-m';
+    protected $format = '!Y-m';
 
     /**
      * Seed attributes


test/Element/MonthTest.php
```diff
@@ -81,4 +82,11 @@ public function testHTML5MonthValidation(string $value, bool $expected): void
         $monthValidator = $inputSpec['validators'][0];
         self::assertEquals($expected, $monthValidator->isValid($value));
     }
+
+    public function testTestDayTimeReset(): void
+    {
+        $element = new MonthElement('foo');
+        $date    = DateTime::createFromFormat($element->getFormat(), '2023-01');
```
Ocramius, line 89: @samuelgmartins this test is fine, but is there anything from `$element` that we can use to fetch a generated date, perhaps? :thinking:

Ocramius (approved):

Ocramius:
I think we should target `3.10.x` here, since this is a bugfix

test/Element/MonthTest.php
internalsystemerror: I don't think so as all we know about the value is the year and month. I think I would prefer to see a check for the 31st/30th if those are the erroneous dates rather than enforcing date 01.

samuelgmartins:
```php
$date    = DateTime::createFromFormat($element->getFormat(), date('Y-m'));
```
That way?

internalsystemerror: I meant manually setting the date to the 31st of a month and then checking if the original issue still exists.

samuelgmartins: @internalsystemerror I think it's not possible because you need to change the date on the server to the 31st.

I tried to set it directly in the DataStep object using the setBaseValue() function, but this is linked to the format, I tried to send the dataTime object already created, even though I couldn't do the test to check exactly the 31st. 
Follows failed test:
```php
use Laminas\Form\Form;
use Laminas\Form\Element;
use Laminas\InputFilter\InputFilter;
use Laminas\InputFilter\Input;
use Laminas\Validator\DateStep;

$form  = new Form();
$element = new Element\Month('month');
$input = new Input();
$input
    ->getValidatorChain()
    ->attach((new DateStep())
        ->setFormat($element->getFormat())
        ->setStep(new \DateInterval("P1M"))
        ->setBaseValue(\DateTime::createFromFormat('Y-m-d', '1970-01-31', new \DateTimeZone('UTC')))
    )
;
$inputFilter = new InputFilter();
$inputFilter->add($input, 'month');
$form->setUseInputFilterDefaults(false);
$form->setInputFilter($inputFilter);
$form->add($element);
$form->setData(['month' => '2000-01']); // 31 day month
var_dump($form->isValid()); // returns false
var_dump($form->getMessages()); // dateStepNotStep
$form->setData(['month' => '2000-04']); // 30 day month
var_dump($form->isValid()); // returns false
var_dump($form->getMessages()); // dateStepNotStep
```
So, as suggested, I ended up checking to see if the date was reset.

internalsystemerror: Yeah I think the problem stems from this part of the DateStep validator https://github.com/laminas/laminas-validator/blob/c5f73b1dc9b657af24736b56318bcd880d9fee94/src/DateStep.php#L254. We might be able to use a psr/clock implementation to help, but that shouldn't hold this PR up.

internalsystemerror (approved):

test/Element/MonthTest.php
internalsystemerror: I wonder now though, if that is a bug with the validator?

samuelgmartins: > I wonder now though, if that is a bug with the validator?

I believe not because the DateStep validator has to validate other elements with full date and time, being an individuality relative to the Month element that validates only the year and month.

Ocramius:
Since I'm releasing `3.11.0` anyway, rolling with what we have here :) 

No point in delaying it further for semantics

PR #219, opened 2023-06-08
Cache `stripos()` call result when checking for delimiters in a loop

pine3ree:
Avoid calling stripos multiple time for the same result


Ocramius (approved):
CI failures unrelated: nice and simple optimization, thanks!

pine3ree:
@Ocramius ...about the QA CI failures: I tried updating the phpunit.xml.dist schema as suggested (`include` not recognized), but it didn't help ( 2 different phpunit versions requiring different schemas?), so I reverted it back to the original file b4 opening the PR

Ocramius:
@pine3ree the solution here is bumping to latest PHPUnit - can be done separately

Issue #221, opened 2023-06-09
keys(option-values) generation in get(Hours|Minutes|Seconds|...)Options


pine3ree:
I have a doubt about using IntlDateFormatter for options keys generations in Date//DateTime Select view helpers. The patterns used are:
```php
'MM' // for months
'dd' // for days
'HH' // for hours
'mm' // for minutes
'ss' // for seconds
```
At first sight one would think that all thos values are zerofilled integer strings, which could be fetched via a simple:
```php 
sprintf('%02d', $value);  // also faster, no formatter instance needed
```

But testing it with various locales we get non numeric symbols. Example:

```php
$datetime = new DateTimeImmutable();

$locales = IntlCalendar::getAvailableLocales();

echo "\n";

foreach ($locales as $locale) {
    $dateFormatter = new IntlDateFormatter(
        $locale,
        IntlDateFormatter::LONG,
        IntlDateFormatter::LONG,
        null,
        null,
        'MM dd HH mm ss' // Combining all the patterns used in the options builders 
    );

    $formattedDate = $dateFormatter->format($datetime);

    if (!preg_match('/^[a-zA-Z0-9\:\-\s]+$/', $formattedDate)) {
        echo "$locale => " . $dateFormatter->format($datetime) . "\n";
    }
}

/*
RESULTS
ar => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_001 => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_BH => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_DJ => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_EG => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_ER => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_IL => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_IQ => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_JO => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_KM => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_KW => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_LB => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
ar_MR => ٠٦ ٠٩ ١٩ ٥٧ ٣٤
[...]

PR #222, opened 2023-06-12
Add validity-tests for form(Date|DateTime|Month)Select formatters option values

pine3ree:
Add PoC tests to prove that current datetime select formatters output non numeric option-values (see issue: https://github.com/laminas/laminas-form/issues/221)

This PR is incomplete as it only add test for month-select option-values. All remaining tests should be added if the alleged bug is recognized to be a real bug. Same goes for related suggested fix (https://github.com/laminas/laminas-form/pull/223)


pine3ree:
Fixed by https://github.com/laminas/laminas-form/pull/230

PR #223, opened 2023-06-12
3.x.x add fix for datetime select option values

pine3ree:
PoC specific fix for test introduced in https://github.com/laminas/laminas-form/pull/222. (see issue https://github.com/laminas/laminas-form/issues/221)

The fix is incomplete. Fixes should be applied to all other selects (day, hour, minute, second).

Differently from other select components, the year select always outputs english digits for years. I believe we should format years option-labels (array-values) as well, while keeping numeric option-values (array-keys)


pine3ree:
Completed by https://github.com/laminas/laminas-form/pull/230

PR #226, opened 2023-06-27
Fix CI Failures with PHPUnit on latest

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Fixes PHPUnit failing on latest deps because the config file doesn't validate against the latest schema.


Ocramius (approved):
Thanks @gsteel!

PR #231, opened 2023-06-30
3.13.x datetime select get elements

pine3ree:
Make sure that `DateTimeSelect::getElements()` returns all its sub `Select` elements


pine3ree:
Q: any reason for not having getElements() return element from less specific to more specific (Y m d H i s) or even better an hash map `['year' => $yearElement, 'month' => $monthElement, ...]` ? kind regards.

Slamdunk (approved):
I don't know the answer, @pine3ree but this looks good

PR #232, opened 2023-07-01
3.13.x intl lowercase m for minute

pine3ree:
The unicode symbol for minute is lowercase 'm' so using `str_contains` or `strpos` is more appropriate

see: https://www.unicode.org/reports/tr35/tr35-dates.html#dfst-minute
see: https://unicode-org.github.io/icu/userguide/format_parse/datetime/#date-field-symbol-table


Slamdunk (approved):
That's even more correct considering the already present `str_contains($value, 'M')` in an if above.

Thank you

PR #235, opened 2023-07-03
3.13.x fix cached pattern bugs

### LINKED ISSUE #233: Date/Time selects helpers __invoke args may have no effect
As explained in the issue title we can call for invoke `FormDateTimeSelect.` with a few arguments:
```php
    public function __invoke(
        ?ElementInterface $element = null,
        int $dateType = IntlDateFormatter::LONG,
        int $timeType = IntlDateFormatter::LONG,
        ?string $locale = null
    ) {
   //...
```

but in doing that `getPattern()` is called internally which result is cached to the internal property `pattern` with the values for dateType, timeType and locale of its first call.
https://github.com/laminas/laminas-form/blob/d20ba468fad89a23e68ed679b6b827543e166afa/src/View/Helper/FormDateTimeSelect.php#L180

We should either remove the cache property (from the base abstract helper) or stale/reset its value to `null` when calling either `setDateType()` or `setTimeType()` or `setLocale()`.

This affects all `AbstractFormDateSelect` descendants, namely: `FormMonthSelect`, `FormDateSelect` and `FormDateTimeSelect`

This also affects calling `render()` after calling `setDateType()` or `setTimeType()` or `setLocale()`, as the internal pattern is cached during the 1st rendering call with the values of `$dateType`, `$timeType`, `$locale` at that time.

pine3ree:
Fix for https://github.com/laminas/laminas-form/issues/233

(We could also move the assignments inside the conditionals, as the assignment makes sense if the value is changed)

### DIFF
--- a/src/View/Helper/AbstractFormDateSelect.php
+++ b/src/View/Helper/AbstractFormDateSelect.php
@@ -123,6 +123,10 @@ abstract class AbstractFormDateSelect extends AbstractHelper
             $dateType = IntlDateFormatter::LONG;
         }
 
+        if ($this->dateType !== $dateType) {
+            $this->pattern = null;
+        }
+
         $this->dateType = $dateType;
 
         return $this;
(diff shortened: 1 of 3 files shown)


Slamdunk (approved):
Thank you very much :ship:

Issue #236, opened 2023-07-03
IntlDateFormatter::SHORT is not handled correctly by date-time select helper


pine3ree:
The helper assumes that any parsed date-time pattern returns  an array with a `second` key:
https://github.com/laminas/laminas-form/blob/d20ba468fad89a23e68ed679b6b827543e166afa/src/View/Helper/FormDateTimeSelect.php#L107

pine3ree:
There are a couple of solution for this:
1. Always use the long format for time (Removing setter and arguments BC) and use `shouldShowSeconds`
2. Set `shouldShowSeconds` to false if `$pattern['second']` is not set, moving all related parts inside the conditional
   https://github.com/laminas/laminas-form/blob/d20ba468fad89a23e68ed679b6b827543e166afa/src/View/Helper/FormDateTimeSelect.php#L132

There is also another wrong assumption:
https://github.com/laminas/laminas-form/blob/d20ba468fad89a23e68ed679b6b827543e166afa/src/View/Helper/FormDateTimeSelect.php#L137
as in some locales the time part comes before the date part, so assuming the index number for a delimiter makes little sense

PR #237, opened 2023-07-03
PoC test showing that short intl-format for time part is not handled

pine3ree:
See issue: https://github.com/laminas/laminas-form/issues/236


Slamdunk:
Solved by https://github.com/laminas/laminas-form/pull/258

PR #245, opened 2023-11-16
Improve type inference for element attributes

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes

### Description

Marks all attribute arrays and iterables as `<string, mixed>` which is slightly better than `<array-key, mixed>`

This is pretty minor stuff, but it really helps in consumer projects running psalm.

I considered that we could have gone for `<string, scalar|null>` but I'm fairly sure there are cases where arrays are accepted, or maybe even iterables at which point, `<string, mixed>` felt like a better middle ground considering `ElementInterface::getAttribute(string $name): mixed`

### DIFF
--- a/src/Element/MultiCheckbox.php
+++ b/src/Element/MultiCheckbox.php
@@ -11,6 +11,10 @@ use Laminas\Validator\ValidatorInterface;
 
 use function assert;
 use function is_array;
+use function is_iterable;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
 
 class MultiCheckbox extends Checkbox
 {
@@ -98,9 +102,13 @@ class MultiCheckbox extends Checkbox
     /** @inheritDoc */
     public function setAttribute(string $key, $value)
     {
-        // Do not include the options in the list of attributes
-        // TODO: Deprecate this
-        if ($key === 'options') {
+        /** @psalm-suppress DocblockTypeContradiction */
+        if ($key === 'options' && is_iterable($value)) {
+            trigger_error(
+                'Providing multi-checkbox value options via attributes is deprecated and will be removed in '
+                . 'version 4.0 of this library',
+                E_USER_DEPRECATED,
+            );
             $this->setValueOptions($value);
             return $this;
         }
--- a/src/Element/MultiCheckbox.php
+++ b/src/Element/MultiCheckbox.php
@@ -14,11 +14,7 @@ use function is_array;
 
 class MultiCheckbox extends Checkbox
 {
-    /**
-     * Seed attributes
-     *
-     * @var array
-     */
+    /** @var array<string, scalar|null>  */
     protected $attributes = [
         'type' => 'multi_checkbox',
     ];
@@ -99,12 +95,7 @@ class MultiCheckbox extends Checkbox
         return $this;
     }
 
-    /**
-     * Set a single element attribute
-     *
-     * @param  mixed  $value
-     * @return $this
-     */
+    /** @inheritDoc */
     public function setAttribute(string $key, $value)
     {
         // Do not include the options in the list of attributes
--- a/src/LabelAwareInterface.php
+++ b/src/LabelAwareInterface.php
@@ -21,7 +21,7 @@ interface LabelAwareInterface
     /**
      * Set the attributes to use with the label
      *
-     * @param  array $labelAttributes
+     * @param  array<string, mixed> $labelAttributes
      * @return $this
      */
     public function setLabelAttributes(array $labelAttributes);
--- a/src/LabelAwareInterface.php
+++ b/src/LabelAwareInterface.php
@@ -21,7 +21,7 @@ interface LabelAwareInterface
     /**
      * Set the attributes to use with the label
      *
-     * @param  array<string, mixed> $labelAttributes
+     * @param  array<string, scalar|null> $labelAttributes
      * @return $this
      */
     public function setLabelAttributes(array $labelAttributes);
@@ -29,7 +29,7 @@ interface LabelAwareInterface
     /**
      * Get the attributes to use with the label
      *
-     * @return array
+     * @return array<string, scalar|null>
      */
     public function getLabelAttributes(): array;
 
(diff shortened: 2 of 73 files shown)


Slamdunk (changes requested):
> I considered that we could have gone for `<string, scalar|null>` but I'm fairly sure there are cases where arrays are accepted

As far as I can test and tell, those cases are errors that come from a past (year 2012) when types weren't a thing, and we tested everything just to be sure but without a valid use case.
I'm talking about:

https://github.com/laminas/laminas-form/blob/943ece1bdb3050b36a964f4aa869d1c9453345b3/test/FactoryTest.php#L99-L105

https://github.com/laminas/laminas-form/blob/943ece1bdb3050b36a964f4aa869d1c9453345b3/test/View/Helper/FormElementTest.php#L87

I'd say let's go all-in with `<string, scalar|null>` even for `ElementInterface::getAttribute` PHPDoc return type, remove the four test cases with invalid sets and release.

src/LabelAwareInterface.php
```diff
@@ -21,7 +21,7 @@ public function getLabel(): ?string;
     /**
      * Set the attributes to use with the label
      *
-     * @param  array $labelAttributes
+     * @param  array<string, mixed> $labelAttributes
```
Slamdunk: I guess we can go `<string, null|scalar>` here too, WDYT?

src/View/Helper/AbstractHelper.php
```diff
@@ -293,7 +293,7 @@ public function getId(ElementInterface $element): ?string
     {
         $id = $element->getAttribute('id');
         if (null !== $id) {
-            return $id;
+            return (string) $id;
```
Slamdunk: Not a fun of type-casting in this case, I'd rather go for an `assert()`

src/View/Helper/FormElement.php
```diff
@@ -20,7 +20,7 @@ class FormElement extends BaseAbstractHelper
     /**
      * Instance map to view helper
      *
-     * @var array
+     * @var array<class-string, string>
```
Slamdunk:
```suggestion
     * @var array<class-string<\Laminas\Form\ElementInterface>, string>
```
And same for `addClass()` param

src/Element/MultiCheckbox.php
```diff
@@ -99,17 +99,16 @@ public function setOptions(iterable $options)
...
     public function setAttribute(string $key, $value)
     {
-        // Do not include the options in the list of attributes
-        // TODO: Deprecate this
-        if ($key === 'options') {
+        /** @psalm-suppress DocblockTypeContradiction */
+        if ($key === 'options' && is_iterable($value)) {
+            trigger_error(
+                'Providing multi-checkbox value options via attributes is deprecated and will be removed in '
+                . 'version 4.0 of this library',
```
Ocramius, line 109: Does building the element from an array spec still work, after removing this?

Slamdunk (changes requested):
Really good work, some minor improvements and we can go :muscle:

Ocramius (approved):
Very cool! One question on the patch, otherwise LGTM!

src/Element/MultiCheckbox.php
Slamdunk: @gsteel seems a prudent guy:

https://github.com/laminas/laminas-form/pull/245/files#diff-9ed265bc15aed45c500eadeea5fec4faa6af32a6800c80a282eac53c51b604d4R188-R216

Ocramius: Was more thinking about the really wonky early-laminas "let's make an entire form from a single array" kinda thing :D

gsteel:
Thanks @Slamdunk - I've adjusted the tests to use `setValueOptions` where appropriate and also properly deprecated providing "value options" in the `options` key to `setAttributes` - Also added deprecated tests to cover the deprecation.

Sorry, the patch is a bit bigger now, but the change to scalar|null of course caused a lot of valid psalm issues that needed fixing

src/Element/MultiCheckbox.php
gsteel: Yes, it does still work - The factory tests _were_ showing this, but I did adjust this test in https://github.com/laminas/laminas-form/pull/245/commits/da4a9f0f8e17adaf20bfbd14d53c590f54768ab8 so that the deprecated version is no longer used.

src/View/Helper/AbstractHelper.php
gsteel: I've changed the conditional here instead to `is_string($id) && $id !== ''`

Slamdunk (approved):

Slamdunk:
What a refreshing patch, thank you @gsteel

gsteel:
Thanks to you too @Slamdunk and @Ocramius for the reviews :)

pavattt:
@Slamdunk Do you have any rules on what changes can go into a minor version and what cannot? Signature change in minor version is not something I'd expect.

froschdesign:
@pavattt 
> Signature change in minor version is not something I'd expect.

Only allowed in major versions. Do you find a problem?

pavattt:
@froschdesign I've upgraded laminas/laminas-form from 3.13.1 to 3.17, and now I have a ton of issues in static analysis I have to fix.

Ocramius:
> @Slamdunk Do you have any rules on what changes can go into a minor version and what cannot? Signature change in minor version is not something I'd expect.

It's a bit of a gray area here, because SA shifts regularly (even with patch versions of psalm and phpstan).

We keep BC on anything that would otherwise require changes in the code for:

 * making it run as before
 * making it "compile" as before, where "compile" is what PHP itself understands as such

For changes such as more precise data structures and types, it is very improbable that you'll get BC on that, as the types can easily shift with tiny changes, even in other packages that declare better types.

My general suggestion when doing a `composer.lock` change that introduces new type errors is to have a baseline file (supported by both PHPStan and Psalm), updating that, and deciding which errors that were introduced are worth fixing immediately.

PR #246, opened 2023-11-19
General type inference improvements

gsteel:
The main goal here was to get inference when iterating over a form or fieldset, both for psalm and in yer IDE - this led to fixing a few more psalm issues here and there.

There are a couple of additions to the baseline occurring due to improved inference, but it would have been unwise to fix these.

### DIFF
--- a/src/Fieldset.php
+++ b/src/Fieldset.php
@@ -15,11 +15,13 @@ use Traversable;
 
 use function array_key_exists;
 use function assert;
-use function gettype;
+use function get_debug_type;
 use function in_array;
 use function is_array;
 use function is_iterable;
+use function is_numeric;
 use function is_object;
+use function is_string;
 use function ltrim;
 use function sprintf;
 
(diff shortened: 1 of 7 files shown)


src/Fieldset.php
```diff
@@ -301,10 +306,7 @@ public function getMessages(?string $elementName = null): array
             $messages = $this->messages;
             foreach ($this->iterator as $name => $element) {
                 $messageSet = $element->getMessages();
-                if (
-                    empty($messageSet)
-                    || (! is_array($messageSet) && ! $messageSet instanceof Traversable)
-                ) {
+                if ($messageSet === []) {
```
gsteel, line 309: `ElementInterface::getMessages()` has a native `array` return type

src/Fieldset.php
```diff
@@ -203,13 +207,14 @@ public function has(string $elementOrFieldset): bool
      */
     public function get(string $elementOrFieldset): ElementInterface
     {
-        if (! $this->has($elementOrFieldset)) {
+        $element = $this->iterator->get($elementOrFieldset);
```
Ocramius, line 210: Is this behaviour change safe to perform? 🤔

Ocramius (approved):

src/Fieldset.php
Slamdunk: Yes it is:

https://github.com/laminas/laminas-form/blob/b83d29b8fbccb7546c18459745a4875153037973/src/Fieldset.php#L194-L197

Slamdunk (approved):

PR #248, opened 2023-11-23
Introduce Form Template to improve inference for processed payloads

### LINKED ISSUE #238: [RFC] Improve type inference for `getData()`
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

With https://github.com/laminas/laminas-inputfilter/pull/91 landing in input-filter, it'll be easier to annotate forms with something like:

```php
/**
 * @psalm-type ValidPayload = array{
 *     field1: int<1, 365>,
 *     field2: non-empty-string,
 *     field3: list<non-empty-string>,
 * }
 * @extends Form<ValidPayload>
 */
final class MyForm extends Form
{
}
```

It's been a while since I used form for anything other than regular associative arrays, but given forms can be bound to objects too - it's also feasible that forms could be annotated to return models too, such as:

```php
/** @extends Form<SomeModel> */
final class MyForm extends Form
{
}
```

The patch in InputFilter helps because users will be able to do something like the following:

```php
/**
 * @psalm-import-type ValidPayload from MyInputFilter
 * @extends Form<ValidPayload>
 */
final class MyForm extends Form
{
    public function __construct(MyInputFilter $inputFilter)
    {
        parent::__construct();
        $this->setInputFilter($inputFilter);
    }
}
```

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes
| RFC           | yes

### Description

Introduce Form Template to improve inference for processed payloads…

This patch should allow users to provide templates for form classes with full inference of the resulting valid payload.

Added an interface method, which we could probably revert considering that `Form::getData()` has decent inference if you explicitly pass a flag, due to the conditional return on the interface.

There are some weird _new_ issues around inheritance and fluent return types which I'm leaving here for now because I have no idea how to fix them, or why they were introduced by these changes. Maybe they are fixed in 5.16, so I'll update there…

Closes #238


gsteel:
Psalm 5.16 doesn't fix the new `LessSpecificReturnStatement` problems. Maybe someone can take a look and see if they have any idea what's going on there.

Slamdunk (approved):
The overall patch looks very good to me.

My only worry is that if we don't fix the bugs/mess on PSalm here, the users could be flooded with invalid errors as well.

@Ocramius do you know how to get PSalm working, or anyone we could ask for help?

gsteel:
<strike>Sussed it, All/most of the fluid return types in `Form` are annotated with `@return $this`, when inherited, psalm infers the return type as `FormSubclass<TFilteredValues>` and `$this` is therefore `LessSpecific`. Returning `self` or `$this<TFilteredValues>`, fixes the issues. I'll update the patch with the latter, because it is more precise</strike>

Ok, I don't actually think that psalm likes `$this<TFilteredValues>` at all. I found that either `self` or `self<TFilteredValues>` solved the issues in very specific places, but wholesale replacement of `$this` yielded many more errors.

I've updated the Static Analysis test case to illustrate template preservation using fluid return types, but not for all possible methods.

Also, I thought that adding an interface method might be a BC break? Happy to revert that if is.

Slamdunk:
> Also, I thought that adding an interface method might be a BC break? Happy to revert that if is.

Heh

Slamdunk (changes requested):
```console
$ docker run --rm -v `pwd`:/app nyholm/roave-bc-check

[BC] ADDED: Method getValidatedPayload() was added to interface Laminas\Form\FormInterface                                                                                                                                                      
1 backwards-incompatible changes detected
```

I always thought it wasn't, but apparently it is a https://github.com/laminas/laminas-form/labels/BC%20Break so yes please, revert it

src/Element.php
```diff
@@ -311,7 +311,7 @@ public function getLabelAttributes(): array
      *
      * Implementation will decide if this will overwrite or merge.
      *
-     * @return $this
+     * @return self
```
samsonasik: The class is not final, so it needs `@return static`, 

see 

self for both parent and child: https://phpstan.org/r/e408ad6a-0bba-43c6-8d71-2005354de06c
static with self only on final: https://phpstan.org/r/5c73379c-5217-4cb4-bc84-a0ab4bbce09b

gsteel: @samsonasik 
I've reverted those specific changes from `$this` to `self`.

`$this` is the correct return type in all cases, but the psalm issues relating to `LessSpecificReturnStatement` just look _really_ odd to me. Can you take a look at the failures and see if you've got any ideas. I'm starting to feel a bit stumped, or there's possibly a psalm bug going on

Slamdunk: Can we try `static<TFilteredValues>`? Does it make sense?

gsteel: I have tried that with no success - also, we can't use that on `Fieldset::setUseAsBaseFieldset()` for example, because it doesn't have templates.

Adding a template to Fieldset etc might work for some of the issues, I could try that, but it's really not gonna work on `LabelAwareInterface` - if feels like some weird inheritance bug but I'm not clever enough to know that for sure.

gsteel: > Adding a template to Fieldset etc might work for some of the issues

This cannot work because `Fieldset` does not provide any way to access the filtered values after validation. `getValue` simply returns null.

gsteel:
@Slamdunk 
Can I baseline the remaining issues please 🙏
I feel like the issues are concerned with the deep inheritance hierarchy where _everything_ extends from `Element`. You'll see that the issues are not applied to a `SomeUserForm` when extending from `Form` but to ancestors such as `Element` and `Fieldset`.

I'd like to try and keep this patch alive because the additional inference provided to consumers would be really useful for a project I'm working on 😁

Slamdunk:
> Can I baseline the remaining issues please 🙏

As long as you give me a reasonable guarantee that users' PSalm runs won't explode with unresolvable errors after this patch, I'm totally ok with this :+1:

gsteel:
To verify that these psalm issues don't crop up for inheritors of `Form` and `Fieldset` etc, I've fixed a load of issues with test assets to remove them from baseline completely.

Thanks @Slamdunk

Slamdunk (approved):

Ocramius:
Worth releasing: triggered a release

PR #253, opened 2024-01-03
Feature: allow attributes for <fieldset> wrapper

### LINKED ISSUE #62: FormRow: allow attributes for <fieldset> wrapper

mimmi20:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

With this PR Attributes will be allowed an the fieldset wrapper for radio- oder multicheckbox-Elements. This PR is related to Issue #62.

### DIFF
--- a/src/View/Helper/FormRow.php
+++ b/src/View/Helper/FormRow.php
@@ -203,35 +203,35 @@ class FormRow extends AbstractHelper
             || $element instanceof MonthSelect
             || $element instanceof Captcha
         ) {
-            $fieldsetHelper = $this->getFieldsetHelper();
-            assert(is_callable($fieldsetHelper));
-
-            $collection = new Collection();
-            $collection->add($element);
-
-            $markup = $fieldsetHelper($collection);
-
-//                $legendAttributesData    = $element->getOption('legend_attributes');
-//                $wrapperAttributesData   = $element->getOption('wrapper_attributes');
-//                $wrapperAttributesString = '';
-//                $legendAttributesString  = '';
+            //            $fieldsetHelper = $this->getFieldsetHelper();
+//            assert(is_callable($fieldsetHelper));
 //
-//                if (is_array($legendAttributesData) && $legendAttributesData !== []) {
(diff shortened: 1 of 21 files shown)


src/View/Helper/FormRow.php
```diff
@@ -186,8 +187,24 @@ public function render(ElementInterface $element, ?string $labelPosition = null)
...
+                }
+
+                if (is_array($wrapperAttributesData) && $wrapperAttributesData !== []) {
+                    $wrapperAttributesString = ' '
+                        . (new FormCollection())->createAttributesString($wrapperAttributesData);
+                }
+
                 $markup = sprintf(
-                    '<fieldset><legend>%s</legend>%s</fieldset>',
+                    '<fieldset%s><legend%s>%s</legend>%s</fieldset>',
```
froschdesign: The problem is that the markup (`fieldset` and `legend`) [is still hard-coded](https://github.com/laminas/laminas-form/issues/62#issue-591668102). The idea was therefore to use the wrapper functionality from the `FormCollection` helper and extend it. This would allow the same handling between the different helpers.
The `legend_attributes` option which is introduced here also specifies that it is always a `legend` element.

froschdesign: In addition, I ask myself if the handling is not too cumbersome for the user, because the same options would have to be set for each MultiCheckbox or Radio element every time. 🤔 
(Related to https://github.com/laminas/laminas-form/issues/62#issuecomment-618213926)

PR #256, opened 2024-01-04
Extract helper functions

mimmi20:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This PR wants to extract the code for translating and escaping labels into 2 helper functions.

### DIFF
--- a/src/View/Helper/AbstractHelper.php
+++ b/src/View/Helper/AbstractHelper.php
@@ -7,6 +7,7 @@ namespace Laminas\Form\View\Helper;
 use Laminas\Escaper\Exception\RuntimeException as EscaperException;
 use Laminas\Form\ElementInterface;
 use Laminas\Form\Exception\InvalidArgumentException;
+use Laminas\Form\LabelAwareInterface;
 use Laminas\I18n\View\Helper\AbstractTranslatorHelper as BaseAbstractHelper;
 use Laminas\View\Helper\Doctype;
 use Laminas\View\Helper\EscapeHtml;
@@ -557,5 +558,33 @@ abstract class AbstractHelper extends BaseAbstractHelper
         }
 
         return false;
+    }
+
+    /**
+     * translate the label
+     */
+    protected function translateLabel(string $label): string
+    {
+        if (null !== ($translator = $this->getTranslator())) {
+            $label = $translator->translate(
+                $label,
+                $this->getTranslatorTextDomain()
+            );
+        }
+
+        return $label;
+    }
+
+    /**
+     * escape the label
+     */
+    protected function escapeLabel(ElementInterface $element, string $label): string
+    {
+        if (! $element instanceof LabelAwareInterface || ! $element->getLabelOption('disable_html_escape')) {
+            $escapeHtmlHelper = $this->getEscapeHtmlHelper();
+            $label            = $escapeHtmlHelper($label);
+        }
+
+        return $label;
     }
 }
--- a/src/View/Helper/AbstractHelper.php
+++ b/src/View/Helper/AbstractHelper.php
@@ -562,6 +562,8 @@ abstract class AbstractHelper extends BaseAbstractHelper
 
     /**
      * translate the label
+     *
+     * @internal
      */
     protected function translateLabel(string|int $label): string|int
     {
@@ -570,6 +572,8 @@ abstract class AbstractHelper extends BaseAbstractHelper
 
     /**
      * escape the label
+     *
+     * @internal
      */
     protected function escapeLabel(ElementInterface $element, string $label): string
     {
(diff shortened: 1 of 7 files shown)


src/View/Helper/AbstractHelper.php
```diff
@@ -558,4 +559,32 @@ protected function hasAllowedPrefix(string $attribute): bool
...
+     */
+    protected function escapeLabel(ElementInterface $element, string $label): string
+    {
+        if (! $element instanceof LabelAwareInterface || ! $element->getLabelOption('disable_html_escape')) {
+            $escapeHtmlHelper = $this->getEscapeHtmlHelper();
+            $label            = $escapeHtmlHelper($label);
+        }
+
+        return $label;
+    }
```
Ocramius: Overall, the refactoring is OK, but promotes more inheritance-based design.

I wonder if there's a way to extract these to non-`protected` API?

src/View/Helper/AbstractHelper.php
```diff
@@ -558,4 +559,32 @@ protected function hasAllowedPrefix(string $attribute): bool
...
+    protected function translateLabel(string $label): string
+    {
+        if (null !== ($translator = $this->getTranslator())) {
+            $label = $translator->translate(
+                $label,
+                $this->getTranslatorTextDomain()
+            );
+        }
+
+        return $label;
```
Ocramius: `return $this->getTranslator()?->translate($label, $this->getTranslatorTextDomain()) ?? $label;` or such? Just wondering if it makes sense, as it potentially adds an always-executed method call :-)

src/View/Helper/AbstractHelper.php
```diff
@@ -558,4 +559,32 @@ protected function hasAllowedPrefix(string $attribute): bool
...
+    }
+
+    /**
+     * escape the label
+     */
+    protected function escapeLabel(ElementInterface $element, string $label): string
+    {
+        if (! $element instanceof LabelAwareInterface || ! $element->getLabelOption('disable_html_escape')) {
+            $escapeHtmlHelper = $this->getEscapeHtmlHelper();
+            $label            = $escapeHtmlHelper($label);
```
Ocramius: Return early, rather than overwriting.

Worth also inverting the conditionals to remove `!` and return early

src/View/Helper/FormSelect.php
```diff
@@ -185,7 +185,7 @@ public function renderOptions(array $options, array $selectedOptions = []): stri
                 $value = $optionSpec['value'];
             }
             if (isset($optionSpec['label'])) {
-                $label = $optionSpec['label'];
+                $label = (string) $optionSpec['label'];
```
Ocramius: The need for added `(string)` casts is a bit suspicious - what happened here?

src/View/Helper/AbstractHelper.php
Ocramius: Actually, it does not call the method, if no translator is set:

```php
<?php

function doesSomethingAndReturns(): string {
    echo 'called';

    return 'haha';
}

$instance = null;

echo $instance?->thisWillNeverBeCalled(doesSomethingAndReturns()) ?? 'alternative';
```

https://3v4l.org/chNbV

src/View/Helper/AbstractHelper.php
mimmi20: Do you mean Traits?

Ocramius: Ha, no, that's even worse :D 

I think this is a decent middle-ground, I'm just trying to see if anything else could be tried :)

src/View/Helper/FormSelect.php
mimmi20: For example the `year` part of the `FormMonthSelect` Element returns an int as value. Where should I cast the value to string? Psalm complains that the Translator dont want an int.

src/View/Helper/AbstractHelper.php
```diff
@@ -558,4 +559,25 @@ protected function hasAllowedPrefix(string $attribute): bool
 
         return false;
     }
+
+    /**
+     * translate the label
+     */
+    protected function translateLabel(string|int $label): string|int
```
Slamdunk, line 568: Let's make these new methods `@internal`, I don't want to add new APIs as `protected` methods in extensible classes are

Slamdunk (changes requested):
Very good, just a small tweak and we can go :muscle:

src/View/Helper/AbstractHelper.php
mimmi20: done

Slamdunk (approved):

Slamdunk:
@mimmi20 would you be so kind to rebase this please?

PR #259, opened 2024-01-09
Reducing the complexity of view helpers

mimmi20:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This PR wants to reduce the complexity of the `FormCollection` and the `FormRow` View Helpers. Also an psalm issue is fixed.

### DIFF
--- a/src/View/Helper/AbstractHelper.php
+++ b/src/View/Helper/AbstractHelper.php
@@ -565,7 +565,7 @@ abstract class AbstractHelper extends BaseAbstractHelper
      *
      * @internal
      */
-    protected function translateLabel(string|int $label): string|int
+    protected function translateLabel(string $label): string
     {
         return $this->getTranslator()?->translate($label, $this->getTranslatorTextDomain()) ?? $label;
     }
--- a/src/View/Helper/Form.php
+++ b/src/View/Helper/Form.php
@@ -64,8 +64,9 @@ class Form extends AbstractHelper
 
         $formContent = '';
 
-        $renderer = $this->getView();
+        $renderer = $this->view;
         assert($renderer instanceof PhpRenderer);
+
         foreach ($form as $element) {
             if ($element instanceof FieldsetInterface) {
                 $formContent .= $renderer->formCollection($element);
--- a/src/View/Helper/Form.php
+++ b/src/View/Helper/Form.php
@@ -66,7 +66,6 @@ class Form extends AbstractHelper
 
         $renderer = $this->getView();
         assert($renderer instanceof PhpRenderer);
-
         foreach ($form as $element) {
             if ($element instanceof FieldsetInterface) {
                 $formContent .= $renderer->formCollection($element);
--- a/src/View/Helper/Form.php
+++ b/src/View/Helper/Form.php
@@ -64,7 +64,7 @@ class Form extends AbstractHelper
 
         $formContent = '';
 
-        $renderer = $this->view;
+        $renderer = $this->getView();
         assert($renderer instanceof PhpRenderer);
 
         foreach ($form as $element) {
--- a/src/View/Helper/FormRow.php
+++ b/src/View/Helper/FormRow.php
@@ -119,14 +119,14 @@ class FormRow extends AbstractHelper
         $elementHelper       = $this->getElementHelper();
         $elementErrorsHelper = $this->getElementErrorsHelper();
 
-        $label           = $element->getLabel();
+        $label           = $element->getLabel() ?? '';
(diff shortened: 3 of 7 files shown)


src/View/Helper/AbstractHelper.php
```diff
@@ -565,7 +565,7 @@ protected function hasAllowedPrefix(string $attribute): bool
      *
      * @internal
      */
-    protected function translateLabel(string|int $label): string|int
+    protected function translateLabel(string $label): string
```
gsteel, line 568: This will fix psalm issues in #260

src/View/Helper/Form.php
```diff
@@ -64,8 +64,9 @@ public function render(FormInterface $form): string
 
         $formContent = '';
 
-        $renderer = $this->getView();
+        $renderer = $this->view;
```
gsteel: Can we guarantee that `$this->view` is not null here? Unless the view is provided to the constructor, I'd say not.

Xerkus: This change is a BC break.

mimmi20: reverted

Xerkus:
@mimmi20 github having incident currently. You might want to wait until problem is resolved.

Slamdunk (approved):
Reviewed with split-diff and hidden-whitespaces: LGTM

src/View/Helper/Form.php
Xerkus: @mimmi20 to explain the BC part. This is a BC break only in a strict sense: this method have a possibility of being overridden in extending classes. I do not see a reason to do so. And impact would be absolutely minuscule if someone is affected. However ` getView()` is a part of helper interface and this components was out for a long time to ensure any kind of unexpected usages could be encountered. Unless it is a part of bug fix or the change has some significant advantage it would be better to avoid it.

src/View/Helper/AbstractHelper.php
Xerkus: Is there a reason int was allowed in the first place?

src/View/Helper/FormRow.php
```diff
@@ -160,81 +160,76 @@ public function render(ElementInterface $element, ?string $labelPosition = null)
...
-                };
-            }
-
-            if ($this->renderErrors) {
-                $markup .= $elementErrors;
-            }
+
+        if ($label === '' || $type === 'hidden') {
+            return $elementString . $elementErrors;
+        }
```
Xerkus, line 166: Just to note.

This and following parts lost conditional for `$this->renderErrors`.
Behavior remains correct because `$elementErrors` is an empty string with render errors checked at line 155 that is not part of the diff.

Xerkus:
@mimmi20 thank you

src/View/Helper/AbstractHelper.php
mimmi20: see https://github.com/laminas/laminas-form/pull/256#discussion_r1442113315

okoram: I think removing 'int' causes a BC Break. For example, if we want to generate pagination and we use only numbers:

    $value = 10;

    /** @var \Laminas\Form\Element\Select $field */
    $field = $this->get('items_per_page');

    $values = [];
    for ($i = 1; $i < 4; $i++) {
        $val = $i * $value;

        $values[$val] = $val;
    }

    $field->setValueOptions($values);

Error message:
> Laminas\Form\View\Helper\AbstractHelper::translateLabel(): Argument # 1 ($label) must be of type string, int given, called in .../laminas/laminas-form/src/View/Helper/FormSelect.php on line 201

https://github.com/laminas/laminas-form/blob/ffdf00a4315b950e47983253a259ad1490ccf6bd/src/View/Helper/FormSelect.php#L201 

The solution is the need to always remember to cast all values to string:

    $values[$val] = (string) $val;

Shouldn't this change be accompanied by a change: https://github.com/laminas/laminas-form/pull/256?

gsteel: I'm with you here. A `(string)` cast should be used for all calls, or `translateLabel` should be altered to accept `scalar` and perform the cast there.

Conversely, all `labels` should be string values regardless, so I do think it's up to the user to convert labels to strings first, but that doesn't stop this from being a BC break based on existing behaviour.

gsteel: See #261

PR #260, opened 2024-01-09
Improve type inference for FormElementManager::get()

gsteel:
Currently, if you provide `FormElementManager::get()` with a FQCN, it infers `ElementInterface` instead of the given FQCN. This patch fixes that…

There are some un-related psalm issues from recent patches. Let me know if you want these baselined here

### DIFF
--- a/src/FormElementManager.php
+++ b/src/FormElementManager.php
@@ -321,9 +321,10 @@ class FormElementManager extends AbstractPluginManager
      * createFromInvokable() will use these and pass them to the instance
      * constructor if not null and a non-empty array.
      *
-     * @param class-string<ElementInterface>|string $name Service name of plugin to retrieve.
+     * @template T of ElementInterface
+     * @param class-string<T>|string $name Service name of plugin to retrieve.
      * @param null|array<mixed> $options Options to use when creating the instance.
-     * @psalm-return ($name is class-string<ElementInterface> ? ElementInterface : mixed)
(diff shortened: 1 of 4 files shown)


Slamdunk (approved):

src/FormElementManager.php
```diff
@@ -321,9 +321,10 @@ public function configure(array $config)
      * createFromInvokable() will use these and pass them to the instance
      * constructor if not null and a non-empty array.
      *
-     * @param class-string<ElementInterface>|string $name Service name of plugin to retrieve.
+     * @template T of ElementInterface
+     * @param class-string<T>|string $name Service name of plugin to retrieve.
      * @param null|array<mixed> $options Options to use when creating the instance.
-     * @psalm-return ($name is class-string<ElementInterface> ? ElementInterface : mixed)
+     * @psalm-return ($name is class-string<T> ? T : mixed)
```
Xerkus:
```suggestion
     * @psalm-return ($name is class-string<ElementInterface> ? T : ElementInterface)
```

This does not work as expected. See https://github.com/laminas/laminas-inputfilter/pull/97#discussion_r1446307372 for more context

Also instead of mixed as fallback return `ElementInterface` as form element manager validates return type.

Add regular `@return` annotation for IDEs

Xerkus (changes requested):
Psalm template constraints are not used in conditional block. See comment

Xerkus (approved):

PR #261, opened 2024-01-11
Fix BC break preventing the use of scalar element labels during render

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

#259 introduces a BC break requiring element labels (and select value options) to be `string`. This patch adds a failing test that passes on previous minors.


Xerkus:
Can non-string even be translated? I feel non-string label should be returned verbatim.

Xerkus (approved):

PR #266, opened 2024-02-01
Helper optional renderer consistency

oldpec:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Inconsistencies in the way different view helpers load plugins meant that while most helpers worked without a renderer being attached on $this->view, a few failed.

This PR brings the few cases where plugins weren't successfully created when $this->view wasn't populated into line with the others and includes unit tests that fail before the change.

- Are you fixing a BC Break?

Arguably?
In some older versions of PHP (ie 7.4) some of these cases (eg FormSelect::getFormHiddenHelper()) would work (because $this->view being null only became a TypeError for method_exists() in PHP 8.0) and then stopped working on current PHP versions.
So these fixes would make code that worked for people under older versions of PHP continue to work as they uplift their PHP and Laminas versions.

There is a little background on:
https://discourse.laminas.dev/t/laminas-form-view-helper-retrieval-consistency/3591

### DIFF
--- a/src/View/Helper/AbstractFormDateSelect.php
+++ b/src/View/Helper/AbstractFormDateSelect.php
@@ -223,11 +223,11 @@ abstract class AbstractFormDateSelect extends AbstractHelper
             return $this->selectHelper;
         }
 
-        if ($this->view && method_exists($this->view, 'plugin')) {
+        if (null !== $this->view && method_exists($this->view, 'plugin')) {
             $this->selectHelper = $this->view->plugin('formselect');
         }
 
-        if (! $this->selectHelper instanceof FormSelect) {
+        if (null === $this->selectHelper) {
             $this->selectHelper = new FormSelect();
         }
 
--- a/src/View/Helper/AbstractFormDateSelect.php
+++ b/src/View/Helper/AbstractFormDateSelect.php
@@ -224,12 +224,15 @@ abstract class AbstractFormDateSelect extends AbstractHelper
             return $this->selectHelper;
         }
 
-        if (method_exists($this->view, 'plugin')) {
+        if ($this->view && method_exists($this->view, 'plugin')) {
             $selectHelper = $this->view->plugin('formselect');
-            assert($selectHelper instanceof FormSelect);
             $this->selectHelper = $selectHelper;
         }
-        assert(null !== $this->selectHelper);
+
+        if (! $this->selectHelper instanceof FormSelect)
+        {
+            $this->selectHelper = new FormSelect();
+        }
 
         return $this->selectHelper;
     }
--- a/src/View/Helper/AbstractFormDateSelect.php
+++ b/src/View/Helper/AbstractFormDateSelect.php
@@ -9,7 +9,6 @@ use IntlDateFormatter;
 use Laminas\Form\Exception;
 use Locale;
 
-use function assert;
 use function extension_loaded;
 use function method_exists;
 use function preg_split;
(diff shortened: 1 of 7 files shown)


src/View/Helper/FormSelect.php
```diff
@@ -298,7 +298,7 @@ protected function renderHiddenElement(SelectElement $element): string
     protected function getFormHiddenHelper(): FormHidden
     {
         if (! $this->formHiddenHelper) {
-            if (method_exists($this->view, 'plugin')) {
+            if ($this->view && method_exists($this->view, 'plugin')) {
```
froschdesign: For consistency, I suggest an early return as is the case in the other helpers:

```php
if ($this->formHiddenHelper !== null) {
    return $this->formHiddenHelper;
} 
```

oldpec: I made minimal changes for each but am happy to bring them more into alignment.
will do null first, ie:
`if (null !== $this->formHiddenHelper) {`
as that seems to be what the others do

src/View/Helper/AbstractFormDateSelect.php
```diff
@@ -224,12 +223,13 @@ protected function getSelectElementHelper(): FormSelect
             return $this->selectHelper;
         }
 
-        if (method_exists($this->view, 'plugin')) {
-            $selectHelper = $this->view->plugin('formselect');
-            assert($selectHelper instanceof FormSelect);
-            $this->selectHelper = $selectHelper;
+        if ($this->view && method_exists($this->view, 'plugin')) {
```
Slamdunk:
```suggestion
        if (null !== $this->view && method_exists($this->view, 'plugin')) {
```

src/View/Helper/AbstractFormDateSelect.php
```diff
@@ -224,12 +223,13 @@ protected function getSelectElementHelper(): FormSelect
...
+            $this->selectHelper = $this->view->plugin('formselect');
+        }
+
+        if (! $this->selectHelper instanceof FormSelect) {
```
Slamdunk:
```suggestion
        if (null === $this->selectHelper) {
```

src/View/Helper/FormMultiCheckbox.php
```diff
@@ -393,7 +393,7 @@ protected function getInputHelper(): FormInput
             return $this->inputHelper;
         }
 
-        if (method_exists($this->view, 'plugin')) {
+        if ($this->view && method_exists($this->view, 'plugin')) {
```
Slamdunk:
```suggestion
        if (null !== $this->view && method_exists($this->view, 'plugin')) {
```

src/View/Helper/FormSelect.php
```diff
@@ -297,14 +297,16 @@ protected function renderHiddenElement(SelectElement $element): string
...
-                $this->formHiddenHelper = $this->view->plugin('formhidden');
-            }
+        if (null !== $this->formHiddenHelper) {
+            return $this->formHiddenHelper;
+        }
 
-            if (! $this->formHiddenHelper instanceof FormHidden) {
-                $this->formHiddenHelper = new FormHidden();
-            }
+        if ($this->view && method_exists($this->view, 'plugin')) {
```
Slamdunk:
```suggestion
        if (null !== $this->view && method_exists($this->view, 'plugin')) {
```

src/View/Helper/FormSelect.php
```diff
@@ -297,14 +297,16 @@ protected function renderHiddenElement(SelectElement $element): string
...
+            $this->formHiddenHelper = $this->view->plugin('formhidden');
+        }
+
+        if (! $this->formHiddenHelper instanceof FormHidden) {
```
Slamdunk:
```suggestion
        if (null === $this->formHiddenHelper) {
```

Slamdunk (changes requested):
Thank you very much for fixing this bug :muscle: 

Just minor improvements: only booleans should be put in `if` condition, and condition should be as narrow as possible to let unknown bugs emerge as soon as possible.

Then we can :ship: it

Slamdunk (approved):

PR #273, opened 2024-06-29
Added novalidate to boolean attributes array

krytenuk:
I have added novalidate to the `AbstractHelper::$booleanAttributes` array as this is a boolean attribute in HTML 5.
Currently setting the `Element::setAttribute('novalidate', true)` in the form when the doc-type is set to HTML5, renders novalidate="1" rather than just novalidate.
https://html.spec.whatwg.org/#attr-fs-novalidate

### DIFF
--- a/src/View/Helper/AbstractHelper.php
+++ b/src/View/Helper/AbstractHelper.php
@@ -51,14 +51,14 @@ abstract class AbstractHelper extends BaseAbstractHelper
      * @var array
      */
     protected $booleanAttributes = [
-        'autofocus' => ['on' => 'autofocus', 'off' => ''], // https://html.spec.whatwg.org/#attr-fe-autofocus
-        'checked'   => ['on' => 'checked',   'off' => ''], // https://html.spec.whatwg.org/#attr-input-checked
-        'disabled'  => ['on' => 'disabled',  'off' => ''], // https://html.spec.whatwg.org/#attr-fe-disabled
-        'itemscope' => ['on' => 'itemscope', 'off' => ''], // https://html.spec.whatwg.org/#attr-itemscope
-        'multiple'  => ['on' => 'multiple',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-multiple
-        'readonly'  => ['on' => 'readonly',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-readonly
-        'required'  => ['on' => 'required',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-required
-        'selected'  => ['on' => 'selected',  'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
+        'autofocus'  => ['on' => 'autofocus', 'off' => ''], // https://html.spec.whatwg.org/#attr-fe-autofocus
+        'checked'    => ['on' => 'checked',   'off' => ''], // https://html.spec.whatwg.org/#attr-input-checked
+        'disabled'   => ['on' => 'disabled',  'off' => ''], // https://html.spec.whatwg.org/#attr-fe-disabled
+        'itemscope'  => ['on' => 'itemscope', 'off' => ''], // https://html.spec.whatwg.org/#attr-itemscope
+        'multiple'   => ['on' => 'multiple',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-multiple
+        'readonly'   => ['on' => 'readonly',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-readonly
+        'required'   => ['on' => 'required',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-required
+        'selected'   => ['on' => 'selected',  'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
         'novalidate' => ['on' => 'novalidate', 'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
     ];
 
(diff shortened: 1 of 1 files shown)


gsteel:
@krytenuk - The `novalidate` attribute is only applicable to `<form>` elements and the view helper for forms already has the `novalidate` attribute listed:

https://github.com/laminas/laminas-form/blob/3.20.x/src/View/Helper/Form.php#L28-L37

krytenuk:
@gsteel, thank you for your response. 
Yes the attribute is set in the form view helper but not as a boolean value, the only place to currently set this is in the abstract view helper. This Without specifying the novalidate attribute as a boolean attribute, https://github.com/krytenuk/laminas-form/blob/fix/add-novalidate-to-boolean-attributes/src/View/Helper/AbstractHelper.php#L62, it does not render correctly.
The boolean attribute's check and rendering was implemented in https://github.com/laminas/laminas-form/pull/108.

Before implementing my fix, setting the novalidate attribute to true,  `Element::setAttribute('novalidate', true)`, renders, `novalidate="1"`.
And setting the attribute to false,  `Element::setAttribute('novalidate', false)`, renders, `novalidate=""`.
Also setting the attribute to null `Element::setAttribute('novalidate', null)` or an empty string, `Element::setAttribute('novalidate', '')` renders the same `novalidate=""`.
As far as I am aware, all these are rendering incorrectly as novalidate is a boolean attribute according to https://html.spec.whatwg.org/#attr-fs-novalidate, please correct me if I am mistaken on this.

After adding the attribute to the `AbstractHelper::$booleanAttributes` array, the novalidate attribute renders correctly.
`Element::setAttribute('novalidate', true)` just renders `novalidate` and `Element::setAttribute('novalidate', false)` causes the novalidate attribute not to render. I believe this should be the expected behaviour.
Using truthy and falsy values also work.

The current logic will prevent the novalidate being rendered on any other form element as the check for this is done, https://github.com/krytenuk/laminas-form/blob/fix/add-novalidate-to-boolean-attributes/src/View/Helper/AbstractHelper.php#L396, using the array you highlighted above.

src/View/Helper/AbstractHelper.php
```diff
@@ -51,14 +51,15 @@ abstract class AbstractHelper extends BaseAbstractHelper
...
-        'selected'  => ['on' => 'selected',  'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
+        'autofocus' =>  ['on' => 'autofocus',  'off' => ''], // https://html.spec.whatwg.org/#attr-fe-autofocus
+        'checked'   =>  ['on' => 'checked',    'off' => ''], // https://html.spec.whatwg.org/#attr-input-checked
+        'disabled'  =>  ['on' => 'disabled',   'off' => ''], // https://html.spec.whatwg.org/#attr-fe-disabled
+        'itemscope' =>  ['on' => 'itemscope',  'off' => ''], // https://html.spec.whatwg.org/#attr-itemscope
+        'multiple'  =>  ['on' => 'multiple',   'off' => ''], // https://html.spec.whatwg.org/#attr-input-multiple
+        'readonly'  =>  ['on' => 'readonly',   'off' => ''], // https://html.spec.whatwg.org/#attr-input-readonly
+        'required'  =>  ['on' => 'required',   'off' => ''], // https://html.spec.whatwg.org/#attr-input-required
+        'selected'  =>  ['on' => 'selected',   'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
+        'novalidate' => ['on' => 'novalidate', 'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
```
froschdesign: Adding `novalidate` here is correct, but everything else must be reverted. See also the automated checks.

froschdesign (changes requested):

src/View/Helper/AbstractHelper.php
krytenuk: @froschdesign  I have reverted all but my change but the automated checks are still failing

gsteel: @krytenuk - The array arrows have to line up as per the coding standard. You can run `vendor/bin/phpcbf` and it will fix it for you. In this case, the BC-Check failure will continue to fail and you can't change that.

src/View/Helper/AbstractHelper.php
```diff
@@ -51,14 +51,15 @@ abstract class AbstractHelper extends BaseAbstractHelper
...
-        'selected'  => ['on' => 'selected',  'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
+        'autofocus'  => ['on' => 'autofocus', 'off' => ''], // https://html.spec.whatwg.org/#attr-fe-autofocus
+        'checked'    => ['on' => 'checked',   'off' => ''], // https://html.spec.whatwg.org/#attr-input-checked
+        'disabled'   => ['on' => 'disabled',  'off' => ''], // https://html.spec.whatwg.org/#attr-fe-disabled
+        'itemscope'  => ['on' => 'itemscope', 'off' => ''], // https://html.spec.whatwg.org/#attr-itemscope
+        'multiple'   => ['on' => 'multiple',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-multiple
+        'readonly'   => ['on' => 'readonly',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-readonly
+        'required'   => ['on' => 'required',  'off' => ''], // https://html.spec.whatwg.org/#attr-input-required
+        'selected'   => ['on' => 'selected',  'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
+        'novalidate' => ['on' => 'novalidate', 'off' => ''], // https://html.spec.whatwg.org/#attr-option-selected
```
gsteel:
```suggestion
        'novalidate' => ['on' => 'novalidate', 'off' => ''], // https://html.spec.whatwg.org/#attr-fs-novalidate
```

gsteel (changes requested):

src/View/Helper/AbstractHelper.php
krytenuk: @gsteel  I tried to run `vendor/bin/phpcbf` locally but it was complaining about `LaminasCodingStandard`. Re-ran `composer install` and now working. Have pushed changes.

Slamdunk (approved):

Slamdunk:
Thank you @krytenuk

PR #283, opened 2025-02-05
Add failing test case for factory creation of elements with required options

gsteel:
It is not possible to use custom elements with required options via configuration, i.e.

```php
$this->add([
  'name' => 'foo',
  'type' => Whatever::class,
  'options' => ['some' => 'thing'],
]);
```

This can be fixed in [Factory](https://github.com/laminas/laminas-form/blob/3.22.x/src/Factory.php#L110) by passing any options found in `$spec` but this reveals more problematic issues with collections, for example. Because the options are given to the Collection at construction time, it creates the nested element(s) before the factory has been injected (via setter injection in the plugin manager initialisers) - This causes a fresh plugin manager to be created inside the collection with a call to `->get('NestedElement')` for an unknown element (By default the plugin manager will auto add an invokable factory).

### DIFF
--- a/test/FactoryTest.php
+++ b/test/FactoryTest.php
@@ -23,6 +23,7 @@ use Laminas\Validator\Digits;
 use Laminas\Validator\ValidatorChain;
 use Laminas\Validator\ValidatorInterface;
 use Laminas\Validator\ValidatorPluginManager;
+use LaminasTest\Form\TestAsset\CustomElementWithRequiredOption;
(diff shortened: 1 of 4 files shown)


gsteel:
Probaly related to #251 and #3

Slamdunk:
I got to be honest I forgot how any of this works: the `FormElementManager` has always been so clumsy and cumbersome that I move to `$form->add(new MyElement);` everywhere long time ago

PR #286, opened 2025-06-04
Upgrade to Psalm v6

gsteel:
- Stops ignoring platform reqs on 8.4
- Bumps dev deps and upgrade Psalm
- (Technical BC Break) `ExceptionInterface` extends `Throwable`
- Mark most classes with soft final in `src`
- Add final keyword to everything in `test`
- Adjust wide range of doc comments to mostly inherit from interfaces
- Return `self` instead of `$this` in fluent interfaces, Psalm seems to have changed its mind
- Drops `laminas-db` from dev deps as no symbols are used from there


gsteel:
Added `mixed` to `Element::setValue()` because it is declared that way on the interface - this shouldn't be a BC break because it doesn't force adding `mixed` on to inheritors IIRC, so not sure why the BC checker rejects this.

Can revert if necessary 👍 

https://3v4l.org/ntSR2

Slamdunk (approved):

gisostallenberg:
@gsteel 

> Return self instead of $this in fluent interfaces, Psalm seems to have changed its mind

This makes that PHPStan detects this as type ElementInterface, while it actually (also) is a FormInterface:
```php
$form = (new Form())->setName('name');
```

I understand you might not support PHPStan, but what actual Psalm change made this change necessary?

At least the return type for this methods that can be (and are) inherited should not be `self`, but be `static`, while I think `$this` suggests the object is the same one, where `@return static` could also mean a new one, as with immutable objects, where method calls return a new instance.

gsteel:
@gisostallenberg

I can't remember TBH, I think `self` is problematic in Psalm on a parent/interface as I don't believe it is narrowed in subclasses/implementations. `$this` should yield "The thing you're calling the method on" in Psalm at least. I think `static` is fine though - if `static` "works" for both Psalm and PHPStan, it would be worth sending in a patch 👍 - None of the classes are `readonly` or `@immutable`, so the semantics of `static` should be fine IMO.

PR #287, opened 2025-06-04
Add a Makefile to improve development experience

gsteel:
- Drops the custom markdown lint config
- Adds composer-require-checker _(with a pretty big baseline)_
- Adds a CI Job to run CRC
- Makefile allows linting the docs tree locally (with docker) and building the docs (also with docker) amongst other useful shortcuts such as `make qa` to run all QA tools

- [x] Docs will need to be fixed up now we're using the default markdown lint config


tools/crc/config.json
```diff
@@ -0,0 +1,58 @@
+{
+    "symbol-whitelist" : [
```
Slamdunk, line 2: :thinking: 

After your round of clean up, I need to fix this

Slamdunk (approved):

gsteel:
@Slamdunk I'll leave it you to release if you want to look at the undeclared deps situation

PR #295, opened 2025-10-14
Update docs-build workflow

Xerkus:
Replaces legacy ssh deploy key method


Xerkus:
CI failure is unrelated, caused by recent inputfilter bugfix

PR #302, opened 2025-11-07
Remove soft final from form elements

gsteel:
Making form elements soft `@final` is problematic. Most usage of Form elements involves abusing inheritance and exactly which elements should be marked final requires further consideration rather than blindly applying a simple blanket policy.

This patch reverses the addition of soft final to form elements that was carelessly done (by me) during the Psalm v6 upgrade.

Because this is against 3.22.x, there are a few minor code-style changes due to rector upgrades that are likely already present in 3.23.x


samsonasik (approved):

Slamdunk:
> Most usage of Form elements involves abusing inheritance

This is exactly the reason we should **keep** `@final` annotation

> and exactly which elements should be marked final requires further consideration

It is impossible to arrive at those further consideration by ouselves.

Let's keep the `@final` annotation, an wait for users to ask us for clarification:

1. if inheritance is used to address a missing functionality, we add that in the base class
2. if instead it's used for tricky results, we won't support that

I'm against this PR

gsteel:
@Slamdunk 

Already had (possibly slack) convo with a user - For example, the `Select` element is a very common target - Users will make a factory/element for say, selecting the country and then re-use that `My\Domain\SelectCountry` across many `Form` instances.

As you know, there are better ways of doing this, for example, injecting the value options via factory configuration, or with `setValueOptions`, but typically, this involves dragging the relevant deps into your Forms constructor. Alternatively a service/factory that can generate elements for specific use-cases with common configuration…

However inheritance works well when your elements are factory based, and, the associated attributes change from form to form, such as:

```php
$this->add([
  'name' => 'country',
  'type' => My\SelectCountry::class,
  'options' => [
    'label' => 'Label 1',
  ],
  'attributes' => [
    'required' => true,
    'autocomplete' => 'foo',
  ],
]);
```

In the options that don't involve inheritance, you are either getting an element instance and calling a bunch of setters to mutate the bits that change, or, dragging in deps to populate `options.value_options` inside the form…

I don't disagree with you, so feel free to close this 👍

Slamdunk:
The points you provide are very valid, I've done it that way myself, but this must stop.

Remember that re-opening classes is always a free move: let's keep them close, and I'll be more than happy to discuss and evolve the current (meh) state-of-the-art once a meaningful issue arises.

froschdesign:
There is a discussion on this topic in our forums: https://discourse.laminas.dev/t/best-way-to-use-composition-instead-of-inheritance-for-laminas-form-elements/3978/2

PR #303, opened 2025-11-07
Improve documentation of Select value options array shape

gsteel:
Documents the array shapes for value options and optgroups to improve type inference/safety for users


Slamdunk (approved):
Approve on trust :wink:

PR #306, opened 2025-12-12
Extract collection methods to interface

### LINKED ISSUE #305: Add CollectionInterface
### Feature Request

Extract `Collection` public methods into an interface

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

I have recently upgraded `laminas-form` on a project with a _lot_ of customisations and got a slew a Psalm errors due to many things being marked `@final`. I'm guessing these will become hard `final` in the next major. Fair enough, I understand the motivation and in most cases can create delegates or re-think what the extended classes are trying to do. Busywork I don't need atm, but hey ho.

The extended `Collection` classes will be more difficult: typically these have methods that do things like set select options after the collection has been populated or automatically set `Item #1`, `Item #2` type labels on generated fieldsets. Creating delegates for these will not work: `Fieldset::populateValues()` checks for `instanceof Collection`.

Would you consider a PR to introduce a `CollectionInterface`?

kynx:
This fixes #305, adding `CollectionInterface`.

I ran into to some `DeprecatedInterface` Psalm errors locally on files I haven't touched. I haven't baselined them - let's see what GHA says.

### DIFF
--- a/src/Element/Collection.php
+++ b/src/Element/Collection.php
@@ -29,6 +29,11 @@ use function sprintf;
 /** @final */
 class Collection extends Fieldset implements CollectionInterface
 {
+    /**
+     * @deprecated Please use the constant CollectionInterface::DEFAULT_TEMPLATE_PLACEHOLDER instead
+     */
+    public const DEFAULT_TEMPLATE_PLACEHOLDER = '__index__';
+
     /** @var array|ArrayAccess */
     protected $object;
 
--- a/src/Element/Collection.php
+++ b/src/Element/Collection.php
@@ -27,13 +27,8 @@ use function max;
 use function sprintf;
 
 /** @final */
-class Collection extends Fieldset
+class Collection extends Fieldset implements CollectionInterface
 {
-    /**
-     * Default template placeholder
-     */
-    public const DEFAULT_TEMPLATE_PLACEHOLDER = '__index__';
-
     /** @var array|ArrayAccess */
     protected $object;
 
--- a/src/Element/Collection.php
+++ b/src/Element/Collection.php
@@ -77,7 +77,7 @@ class Collection extends Fieldset implements CollectionInterface
      *
      * @var string
      */
-    protected $templatePlaceholder = self::DEFAULT_TEMPLATE_PLACEHOLDER;
+    protected $templatePlaceholder = CollectionInterface::DEFAULT_TEMPLATE_PLACEHOLDER;
 
     /**
      * Whether or not to create new objects during modify
--- a/src/Element/CollectionInterface.php
+++ b/src/Element/CollectionInterface.php
@@ -7,7 +7,6 @@ namespace Laminas\Form\Element;
 use Laminas\Form\ElementInterface;
 use Laminas\Form\FieldsetInterface;
 
-/** @final */
 interface CollectionInterface extends FieldsetInterface
 {
     /**
--- a/src/Element/CollectionInterface.php
+++ b/src/Element/CollectionInterface.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Form\Element;
+
+use Laminas\Form\ElementInterface;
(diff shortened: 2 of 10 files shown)


kynx:
Huh, I don't think the change the BC check is complaining about actually _is_ a BC break?

src/Element/Collection.php
```diff
@@ -27,13 +27,8 @@
...
 
 /** @final */
-class Collection extends Fieldset
+class Collection extends Fieldset implements CollectionInterface
 {
-    /**
-     * Default template placeholder
-     */
-    public const DEFAULT_TEMPLATE_PLACEHOLDER = '__index__';
-
```
gsteel, line 36: You can't remove this constant, but placing it on the interface is correct. Please reinstate it and add

`/** @deprecated Please use the constant CollectionInterface::DEFAULT_TEMPLATE_PLACEHOLDER  instead */`

src/View/Helper/FormCollection.php
```diff
@@ -170,7 +170,7 @@ public function render(ElementInterface $element): string
     /**
      * Only render a template
      */
-    public function renderTemplate(CollectionElement $collection): string
+    public function renderTemplate(CollectionInterface $collection): string
```
gsteel: This is a BC break for anyone who overrides this method, because this class is not final

src/View/Helper/FormElement.php
```diff
@@ -24,13 +24,13 @@ class FormElement extends BaseAbstractHelper
...
-        Element\DateTimeSelect::class => 'formdatetimeselect',
-        Element\DateSelect::class     => 'formdateselect',
-        Element\MonthSelect::class    => 'formmonthselect',
+        Element\Button::class              => 'formbutton',
+        Element\Captcha::class             => 'formcaptcha',
+        Element\Csrf::class                => 'formhidden',
+        Element\CollectionInterface::class => 'formcollection',
+        Element\DateTimeSelect::class      => 'formdatetimeselect',
+        Element\DateSelect::class          => 'formdateselect',
+        Element\MonthSelect::class         => 'formmonthselect',
```
gsteel, line 34: Adding `CollectionInterface` to this list is fine, but you can't remove `Collection::class`

gsteel (changes requested):
Thanks @kynx - just noting a few BC breaks here

src/View/Helper/FormCollection.php
kynx: Ah OK. I thought because it was `@final` that would be OK.

gsteel: Yeah - soft `@final` is effectively advisory - we can change it for the next major 👍

gsteel (approved):
LGTM 👍

kynx:
Agh. I've got to go - I'll baseline the psalm deprecation warning tomorrow.

gsteel:
> Agh. I've got to go - I'll baseline the psalm deprecation warning tomorrow.

Didn't spot it either - don't baseline it, just change the constant usage to reference the interface instead

kynx:
@gsteel That's done 😄

src/Element/CollectionInterface.php
```diff
@@ -0,0 +1,59 @@
...
+
+declare(strict_types=1);
+
+namespace Laminas\Form\Element;
+
+use Laminas\Form\ElementInterface;
+use Laminas\Form\FieldsetInterface;
+
+/** @final */
+interface CollectionInterface extends FieldsetInterface
```
Slamdunk: A `final` interface ❓

Slamdunk (changes requested):

src/Element/CollectionInterface.php
kynx: Heh, well spotted 😄

Slamdunk: 6 years ago I predicted this: https://x.com/slamzoe/status/1113704907490635776

Slamdunk (approved):

PR #309, opened 2025-12-17
Fix error validating form with nested custom fieldset

kynx:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

If you add a nested fieldset (implementing `FieldsetInterface`) to a form, validation fails with:
```
1) LaminasTest\Form\FormTest::testFormWithNestedCustomCollectionAndValidationGroupBindValuesToModel
TypeError: Laminas\Form\Form::prepareValidationGroup(): Argument #1 ($formOrFieldset) must be of type Laminas\Form\Fieldset, MockObject_CollectionInterface_7c6d4c8a given, called in /Users/matt/www/laminas-form/src/Form.php on line 566
```

This change allows nested `FieldsetsetInterfaces` to be validated.

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -14,7 +14,6 @@ use Laminas\InputFilter\InputFilterProviderInterface;
 use Laminas\InputFilter\InputProviderInterface;
 use Laminas\InputFilter\ReplaceableInputInterface;
 use Laminas\Stdlib\ArrayUtils;
-use Laminas\Stdlib\PriorityList;
 use Traversable;
 
 use function array_key_exists;
@@ -23,6 +22,7 @@ use function assert;
 use function in_array;
 use function is_array;
 use function is_object;
+use function iterator_to_array;
 use function sprintf;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Form.php
```diff
@@ -534,10 +535,15 @@ public function getValidationGroup(): ?array
...
-    {
+    protected function prepareValidationGroup(
+        FieldsetInterface $formOrFieldset,
+        array $data,
+        array &$validationGroup
+    ): void {
         foreach ($validationGroup as $key => &$value) {
-            $fieldset = $formOrFieldset->iterator->get((string) $key);
+            $iterator = $formOrFieldset->getIterator();
+            assert($iterator instanceof PriorityList);
```
kynx: Not 100% sure what to do here. `FieldsetInterface` extends `IteratorAggregate`, which returns a `Traversable` from `getIterator()`. `Fieldset::getIterator()` narrows that to `PriorityList`.

Should `FieldsetInterface` narrow the return as well? I guess that would be a BC break?

kynx: FWIW, this method _used_ to accept `FieldsetInterface`. It was changed here: https://github.com/laminas/laminas-form/commit/d30e93810ee33ad4c24bfdac97a462119604f1cd#diff-92886d01da016b7104e7ce311d6bed7236b2b6dfae8e378f9a4e53d568b8aae1, apparently to make Psalm happy.

Slamdunk (approved):

Slamdunk:
Thank you @kynx

PR #327, opened 2026-09-19
Use StructArmed on QA

samsonasik:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

This PR to make use of [StructArmed](https://github.com/boundwize/structarmed).

This PR start apply StructArmed with initialize using `PER` ( extended of Psr12 extended of psr1 extended of psr4)  and `CODEQUALITY` preset, and already help fix 4 violations.

```diff
```diff
diff --git a/test/Annotation/AbstractBuilderTestCase.php b/test/Annotation/AbstractBuilderTestCase.php
index dc68a38d..ce1d624e 100644
--- a/test/Annotation/AbstractBuilderTestCase.php
+++ b/test/Annotation/AbstractBuilderTestCase.php
@@ -577,7 +577,7 @@ abstract class AbstractBuilderTestCase extends TestCase
         $entity  = new TestAsset\Annotation\Entity();
         $builder = $this->createBuilder();
 
-        $builder->getEventManager()->attach('*', function () {
+        $builder->getEventManager()->attach('*', static function () {
             // return void;
         });
 
diff --git a/test/Annotation/BuilderAbstractFactoryTest.php b/test/Annotation/BuilderAbstractFactoryTest.php
index ebd1fab8..de521917 100644
--- a/test/Annotation/BuilderAbstractFactoryTest.php
+++ b/test/Annotation/BuilderAbstractFactoryTest.php
@@ -174,7 +174,7 @@ final class BuilderAbstractFactoryTest extends TestCase
                 ['config', $config],
                 [
                     'test-listener',
-                    new class () {
+                    new class {
```

to use `static` on anonymous function when possible and that anonymous class without arguments should no parentheses https://www.php-fig.org/per/coding-style/#8-anonymous-classes.

applied with single command:

```bash
vendor/bin/structarmed analyze --fix
```

I also enable ruleset above so it show more ability of layering setup to define which layers each layer is allowed to depend on.


samsonasik:
I've updated test ICU seems due to latest update 

- https://github.com/unicode-org/cldr/pull/3244

samsonasik:
I also fix psalm notice 👍

samsonasik:
Ready for review 👍

Slamdunk (commented):
Note to self: tool enforced by https://github.com/laminas/laminas-ci-matrix-action/pull/414

I have no experience with `StructArmed` so the review must be done by someone else, @gsteel maybe?

samsonasik:
@Slamdunk it already applied on other components, eg: `laminas-diactoros`:

- https://github.com/laminas/laminas-diactoros/pull/232

gsteel (approved):
Thanks @samsonasik LGTM 👍

